PR #9, opened 2016-08-08
Add Sass rem function.

dcmouyard:
Add a Sass rem function, similar to the exising em function.


coreylafferty:
This function still converts the font to ems.  Shouldn't line 11 there be: 

`@return ($px / $base) * 1rem;`

PR #23, opened 2016-08-25
added default menu template and updated main menu

coreylafferty:
This adds a default menu template and updates the main menu template to include expanded, collapsed and active trail classes.


dcmouyard:
I updated the menu templates to more closely match the markup we used for Drupal 7. I also put the machine name of the menu as a class, so we no longer need the menu--main.html.twig file.

I also changed nav--main-menu to nav--main sass to match the new class in the template.

In a future pull request I’ll add a theme setting to show/hide the expand/collapse classes like we had in Drupal 7.

PR #32, opened 2016-09-27
Refactor sass

dcmouyard:
Disclaimer: This is a big change that will require some discussion to make sure everyone is on board. :)

I refactored the partial structure a bit to incorporate some ideas from [ITCSS](http://www.creativebloq.com/web-design/manage-large-css-projects-itcss-101517528). I separated the helper classes into extendable versions in base/extendables and classed versions in utilities, which use !important to make sure they override previous styles.

The general idea behind ITCSS to organize your styles from least specific to most specific. It works well within the SMACSS framework, and simply adds more structure to things SMACSS doesn’t cover.

I also borrowed Harry Robert’s [BEM naming convention](http://csswizardry.com/2015/08/bemit-taking-the-bem-naming-convention-a-step-further/) of using namespaces for the different layers (except for the components layer) so layout classes are prefixed with `l-` and utility classes are prefixed with `u-`.

FYI - a good example to see ITCSS in action is in [inuitcss](https://github.com/inuitcss).


dcmouyard:
I’m not sure namespacing the utility classes with `u-` is worth it, since we’d have to override a ton of core Drupal markup (e.g., replacing `visually-hidden` with `u-visually-hidden`.)

We could add the non-namespaced classes that are used by Drupal, but I’m not sure it’s worth it.

coreylafferty:
@dcmouyard I made some small tweaks to this:

- Added an _overrides.scss file for minor tweaks to modules (e.g., the now included _toolbar.scss)
- Added an empty /overrides/ directory for complete css file overrides
- Moved nav extendable styles back to _nav.scss and enabled subfolder sass globbing in /components/
- set path for boxsizing.htc to default install path (altho we'll probably remove this altogether as part of getting rid of IE support)
- reverted accessibility classes to not use 'u-' so we don't need to change core Drupal classes (can revisit this if need be)

coreylafferty:
Reassigning to myself to update templates to match the class name changes in this PR.

PR #33, opened 2016-09-28
Add x-ua-compatible meta tag

dcmouyard:
Added `<meta http-equiv="x-ua-compatible" content="ie=edge">` to the markup.


coreylafferty:
Would there be drawbacks to hardcoding this into the html.html.twig file? I know we discussed it, but can't remember where we landed.

PR #36, opened 2016-09-28
Update region template

dcmouyard:


templates/layout/region.html.twig
```diff
@@ -2,26 +2,15 @@
...
  * @file
  * Theme override to display a region.
- *
- * Available variables:
- * - content: The content for this region, typically blocks.
- * - attributes: HTML attributes for the region div.
- * - region: The name of the region variable as defined in the theme's
- *   .info.yml file.
- *
- * @see template_preprocess_region()
```
fafnirical, line 12: Why was this removed from the comment? It's still valid.

dcmouyard: We typically removed all of the documentation comments within the template files in D7 version of Gesso, so I started doing that for the D8 version as well.

I think we originally removed them because they weren’t always accurate and I liked seeing the code closer to the top rather than scrolling through the long comments. :)

Should we revisit that approach for the D8 version? @coreylafferty @fafnirical

fafnirical: I don't have any particularly strong opinions on it, but let's create an issue to discuss?

EDIT: Created https://github.com/forumone/gesso/issues/37 for discussion.

PR #40, opened 2016-11-29
updated field template to print plain content if no label and cleaner…

coreylafferty:
… usage of field__items wrapper

### DIFF
--- a/templates/field/field.html.twig
+++ b/templates/field/field.html.twig
@@ -53,20 +53,9 @@
 %}
 
 {% if label_hidden %}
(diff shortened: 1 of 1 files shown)


dcmouyard:
I don’t think we want any wrapper markup if the label is hidden, even if there are multiple items. Imagine a content field that outputs a bunch of paragraph types. I don’t think we want the field wrapper markup around that.

coreylafferty:
@dcmouyard Sounds good to me.  Make it so.

PR #43, opened 2016-12-02
Update pagers

coreylafferty:


coreylafferty:
Once PR 32 is merged and font variable names finalized , probably a good idea to change the font size in this Sass file to be a variable (e.g., $font-size-s)

dcmouyard:
I broke out the pager styles into its own component that is only loaded when the pager templates are used. I also updated the default styles to more closely follow the U.S. Web Design Standards.

dcmouyard:
@coreylafferty Ready for your review. I moved the component stylesheets to the root sass folder and removed the spaces in the pager template.

PR #48, opened 2016-12-06
Update buttons

dcmouyard:
This pull request depends on #46 being merged first.


dcmouyard:
I updated the default button styles to match the [U.S. Web Design Standards](https://standards.usa.gov/colors/).

This pull request depends on #51 being merged first.

dcmouyard:
This pull request now depends on #46, #51, and #52 being merged first.

sass/partials/components/_button.scss
```diff
@@ -3,15 +3,62 @@
...
       [dir='rtl'] & {
         margin-left: 0;
-        margin-right: 1em;
+        margin-right: 0.5em;
       }
     }
   }
 }
+
+.button--danger {
```
coreylafferty: Do we need all of these extra button styles?  Are they just examples or are any of these classes actually output in Drupal?

coreylafferty (changes requested):

sass/partials/components/_button.scss
dcmouyard: The .button and .button--danger classes are the only ones that Drupal 8 outputs by default. The others are just examples based on what I’ve seen on projects the most. The designs for these button all come from https://standards.usa.gov/buttons/

coreylafferty: Can you add comments for the variants to that effect?  e.g., 

// Default Drupal button class output during the zombie apocalypse.
.button--danger {}

// Custom button class, included as an example. Not a default Drupal class name.
.button--secondary {}

// Size variations, included as examples.  Not default Drupal class names.
.button--lg {}
.button--sm {}

coreylafferty (changes requested):

dcmouyard:
I broke out the button variants to a sub-folder and added comments explaining what they’re for.

coreylafferty (approved):

PR #52, opened 2016-12-08
Update typography

dcmouyard:
I updated the default typography styles to match the [U.S. Web Design Standards](https://standards.usa.gov/typography/).

This pull request depends on #46 and #51 being merged first.


sass/partials/global/variables/_typography.scss
```diff
@@ -2,9 +2,10 @@
 // Typography variables
 // http://sass-lang.com/docs/yardoc/file.SASS_REFERENCE.html#variables_
 
-$base-font-size: 16px;
-$base-line-height: 24px;
+$base-font-size: 17px;
+$base-line-height: 25.5px;
```
coreylafferty: Why 25.5?

sass/partials/global/variables/_typography.scss
```diff
@@ -2,9 +2,10 @@
...
 $base-line-height-ratio: $base-line-height / $base-font-size;
+$vertical-spacing: $base-font-size;
```
coreylafferty: Could a comment be added to explain what this variable is and how it's used?

coreylafferty (changes requested):

sass/partials/global/variables/_typography.scss
dcmouyard: 25.5px for the line-height is just what the U.S. Web Design Standards use, which is the same as a line height of 1.5 for the body copy.

sass/partials/global/variables/_typography.scss
dcmouyard: Good idea. I’ll add a comment and push up a commit later.

sass/partials/global/variables/_typography.scss
coreylafferty: Looking at the guidelines, they appear to set it 26px.  https://standards.usa.gov/typography/

Can we do that, or just set it to a unitless 1.5?

Also, do we need $base-line-height-ratio as a variable?  I don't see it being used (unless it's in another PR).

coreylafferty (changes requested):

dcmouyard:
It turns out the html line-height was supposed to use the $base-line-height-ratio variable so that it’s unit-less.

The $base-line-height variable was often used as the default vertical spacing between elements, and we sometimes used multiples or half of that variable, so we needed a px-unit version of that variable. You can see an example in messages.scss.

Now that we have the vertical-spacing variable, though, perhaps we can remove the base-line-height-ratio variable and change the base-line-height variable to be unit-less. Then we could update messages.scss to use the vertical-spacing variable.

coreylafferty:
That sounds good to me.

dcmouyard:
I refactored the base-line-height usage.

PR #58, opened 2016-12-13
added form-item__url to field list

coreylafferty:
While working on a site, saw there's a field-item__url class created on link fields.  added to the styling for basic fields.


dcmouyard:
@coreylafferty I added color, month, and week form item types that also typically display as a text box.

PR #62, opened 2017-01-12
updated menu classes and added mobile menu js to global library

coreylafferty:


templates/navigation/menu.html.twig
```diff
@@ -25,7 +25,7 @@
         %}
         <ul{{ attributes.addClass(menu_classes) }}>
       {% else %}
-        <ul class="nav__subnav">
```
dcmouyard, line 28: Isn’t the subnav a sub-component of the nav component? I can’t imagine that I would ever use .nav--subnav as a standalone variant. I would always have that nested under the parent menu component.

coreylafferty: The Sass and JS were both looking for nav--subnav, so seemed easiest to just change this one thing to match everything else to get it working again, but doesn't matter to me. If going the other direction, just have to track down all the places it needs to change.

PR #67, opened 2017-01-26
Add mixin to generate pseudo block elements

sbrack8t:


dcmouyard:
I’m not sure if I would ever use this mixin, but if others find this helpful then I have no problems merging it in.

If we do add this, though, should we update any code that uses ::before/::after to use this mixin?

PR #69, opened 2017-01-27
Add mixin to layout constrain with conditional padding

sbrack8t:


sass/partials/layout/_constrain.scss
```diff
@@ -1,15 +1,25 @@
 // @file
 // Styles for constraining the page width.
 
-.l-constrain {
+@mixin constrain($width: em($max-page-width), $padding: true) {
```
dcmouyard: I like the idea of adding a mixin for the constrain styles with optional side padding. I think this should go in global/helpers/_constrain.scss instead of in this file though.

sass/partials/layout/_constrain.scss
```diff
@@ -1,15 +1,25 @@
...
+    padding-left: $gutter-width / 2;
+    padding-right: $gutter-width / 2;
+  }
+}
+
+.l-constrain {
+  @include constrain;
+}
+
+.l-constrain--global {
```
dcmouyard: I don’t see the benefit of adding this variant into Gesso core. Is there a use-case I’m missing that wouldn’t be project-specific?

sass/partials/layout/_constrain.scss
```diff
@@ -1,15 +1,25 @@
...
+  @include constrain(em(2500px), false);
 }
 
 .l-constrain--no-padding {
-  padding-left: 0;
-  padding-right: 0;
+  @include constrain(em($max-page-width), false);
```
dcmouyard: I think this variant should only have the padding-left/right properties since that’s the only thing this variant overrides.

The only reason to add all of the other styles would be to allow people to only add the l-constrain--no-padding class to an element instead of class="l-constrain l-constrain--no-padding".

dcmouyard (changes requested):

sass/partials/layout/_constrain.scss
sbrack8t: I've used this for limiting the entire page width for super large screens

sass/partials/layout/_constrain.scss
sbrack8t: Good point.

sass/partials/layout/_constrain.scss
dcmouyard: Ah, okay. 

Would it make sense to have this width be a variable? Also, we would probably only want to add the max-width property since everything else would already be in l-constrain.

sbrack8t: Makes sense to me.

sbrack8t: But we still need the margin auto to center it on the screen

dcmouyard: It would already have margin auto from l-constrain.

dcmouyard (approved):

dcmouyard:
I moved the classes back into the layout partial, but kept the mixins in global so they could be used anywhere else if needed. 

I turned the l-constrain--break variant into a sub-component (l-constrain__break) since it would always be inside an l-constrain parent. This also allow us to easily override the styles based on the l-constrain variant.

PR #107, opened 2017-11-03
8.x 2.x experimental yo

juranosaurustechs:
Ready for review @coreylafferty


package.json
```diff
@@ -26,9 +26,12 @@
...
     "load-grunt-tasks": "^3.2.0",
-    "postcss-assets": "~4.1.0"
+    "postcss-assets": "~4.1.0",
+    "lodash.includes": "^4.1.3",
+    "yeoman-generator": "^0.23.3",
+    "yo": "^1.8.5"
   },
   "scripts": {
-    "postinstall": "grunt gessoSetup"
+    "postinstall": "grunt gessoSetup", "new-component": "yo ./pattern-lab/source/component-generator/index.js"
```
fafnirical: Let's move this additional script to a separate line

fafnirical: Also: should we be calling `yo` directly here, or using `grunt newComponent`?

_starter-kit/component-generator/index.js
```diff
@@ -0,0 +1,108 @@
+'use strict';
+
+var yeoman = require('yeoman-generator');
+var includes = require('lodash.includes');
+var path = require('path');
+var fs = require('fs');
+var chalk = require('chalk');
+var plBase = ('./pattern-lab/source/_patterns');
```
fafnirical, line 8: Is something missing here? The parenthesis wrapping combined with the previous few lines makes it seem like this line might be missing a `require` call (though I don't believe it is?)

coreylafferty: Since this is just a path, I don't think needs the require line or any changes.

coreylafferty:
Closing due to new work in https://github.com/forumone/gesso/pull/136

PR #110, opened 2017-11-09
[typo-correction] Fix 'patterlab' instances.

lrobeson:
For some reason it's not picking up my _starter-kit corrections, but the same fix needs to be applied to instances in that directory.


coreylafferty:
@lrobeson I only see changes in the _starter-kit folder for this PR.  What changes are you referring to?

PR #125, opened 2018-01-18
update drush file to allow for theme location, whether root, custom o…

chazchumley:
…r contirb folder within themes directory


dcmouyard:
Works like a charm!

PR #127, opened 2018-01-23
update list mixin to allow list-clean to be block level, not inline. …

chazchumley:
… Updated menu.yml to account for original_link attributes


chazchumley:
Backed out block display for list-clean and removed list-inline from _menu.scss and replaced with list-clean

PR #165, opened 2018-08-14
160 refactored files to hardcode static components

coreylafferty:


dcmouyard (approved):
Everything looks good to me.

The only minor nitpick I have is that not all files have a empty line break at the end.

PR #167, opened 2018-10-03
added isset wrapper around required var in form.inc

coreylafferty:


coreylafferty:
Ran into an error (Undefined index #required in form.inc) on a page when adding a field_group wrapper set to create tabs.  Wrapping required in an isset like the other items seems to fix this.

dcmouyard (approved):
Looks good to me.

PR #181, opened 2018-12-11
added |raw where needed and removed if has_? calls

coreylafferty:
Adds raw filter to most printed variables.

Removes has_[variablename] variables and just uses [variablename] instead.


dcmouyard (approved):
Everything looks good.

PR #189, opened 2019-02-20
Attributes update code formatting

dcmouyard:
I updated files to match coding standards for the following minor items:
* Spaces
* 80 character limit on long text in certain places (e.g., README)
* Newline ending on files


_starter-kit/_patterns/03-components/block/block.md
```diff
@@ -10,7 +10,6 @@ _Component for Block_.
 * constrain_classes: [string] Constrain classes used to control layout.
 * label_element: [string] HTML element to use for block label.
 * label: [string] The configured label of the block if visible.
-* title_prefix: [array] Additional output populated by modules, intended to be displayed in front of the main title tag that appears in the template.
-* title_suffix: [array] Additional output populated by modules, intended to be displayed after the main title tag that appears in the template.
+* title_prefix: [array] Additional output populated by modules, intended to beßdisplayed in front of the main title tag that appears in the template.
```
coreylafferty: looks like something happened here

dcmouyard: Hmm, I wonder how that happened? Thanks for catching!

I added the suffix part back in and tweaked the prefix text to say “before” instead of “in front of”.

PR #187, opened 2019-02-12
Gesso config update

sbrack8t:

PR #190, opened 2019-02-26
Add Stylelint

dcmouyard:
This work is based on the Stylelint config Shawn did in #187.

I had to tweak the regex for class/mixin/placeholder names to allow for numbers (e.g., .class-1 now passes).

I also fixed all of the Stylelint errors.

I did _not_ include work to run Prettier automatically on commit or add the Prettier fixes. I’ll create a new PR for that work.


dcmouyard:
We’re not going to add this linting work to the 8.x-2.x branch. #203 is for adding this into the 8.x-3.x branch.

PR #191, opened 2019-02-27
Add Prettier

dcmouyard:
This work adds a pre-commit hook to run Prettier on all staged files. 

I also added all of files changed by Prettier. These changes _shouldn’t_ affect how everything works, but we definitely need to verify. One thing I noticed is that Prettier will wrap commented lines in Yaml files without commenting out the wrapped content, so I had to manually change that in one place.


package.json
```diff
@@ -34,7 +36,18 @@
...
+    "postinstall": "grunt gessoSetup"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{js,json,md,prettierrc,scss,yml}": [
+      "node ./node_modules/.bin/prettier --write",
```
fafnirical, line 49: I don't think `node ./node_modules/.bin/` is required here -- `lint-staged` adds locally-installed Node binaries to $PATH when it executes: https://github.com/okonet/lint-staged#what-commands-are-supported

```suggestion
      "prettier --write",
```

dcmouyard: That’s what I had originally, but it didn’t work. On one of the Github issues I found someone from the Prettier team recommend using `
node ./node_modules/.bin/prettier` instead.

dcmouyard:
Looks like the Prettier fixes breaks the Sass globbing task.

dcmouyard:
I fixed the issues where Prettier was breaking some Grunt and Pattern Lab tasks.

dcmouyard:
After our meeting with Alex on the new Docker container setup, we need to refactor the automated Prettier fixes to run within the linting task instead of using a git commit hook.

PR #193, opened 2019-03-19
added variable for pattern lab

coreylafferty:
Adds a variable for Pattern Lab.  Currently not used in the Drupal version (but is in the WordPress one).  Could be used if there is code that needs to be segmented off for only Drupal or only Pattern Lab.


dcmouyard (approved):
Looks good, this doesn’t affect anything in the Drupal version.

PR #197, opened 2019-04-08
adjusted watch tasks to disable spawning to speed up compile and watch

coreylafferty:


dcmouyard:
This work looks good, but merging 8.x-2.x into the styleint branch introduced a bunch of stylelint errors. I want to clean that up before this gets merged.

dcmouyard (approved):
Nevermind, those errors were from an old version of files in the pattern-lab directory.

PR #200, opened 2019-04-29
Removing raw filter

coreylafferty:


dcmouyard (approved):
Everything looks good and I confirmed that this fix works in a Vagrant setup.

PR #222, opened 2019-08-05
Update breadcrumb pattern

sbrack8t:


source/_patterns/04-components/breadcrumb/_breadcrumb.scss
```diff
@@ -3,10 +3,10 @@
 
 .breadcrumb {
   @include clearfix();
-  background-color: rgba(gesso-palette(blue, dark-1), 0.85);
+  background-color: rgba(gesso-palette(blue, dark-1), 0.9);
```
coreylafferty: for simplicity, how about dropping the rgba() call and just leave it as blue/dark-1?

PR #224, opened 2019-08-08
8.x 3.x config corey

coreylafferty:


source/_patterns/03-layouts/regions/breadcrumb/breadcrumb.twig
```diff
@@ -1,5 +1,7 @@
 <div class="l-breadcrumb">
-  <div class="l-constrain">
+  {# l-constrain commented out to accomodate full-bleed default breadcrumb
```
sbrack8t, line 2: This seems overkill. Should we just create a parent component where we can enable and disable the constrain wrapper? Then all the regions can use it.

source/_patterns/04-components/fieldset/_fieldset.scss
```diff
@@ -50,13 +50,13 @@ $fieldset-legend-offset: rem(gesso-line-height(base) * gesso-base-font-size()) !
 
 .fieldset--default {
   background: #fff;
-  border: 1px solid $fieldset-color-border;
+  border: 1px solid gesso-color(form,border);
```
sbrack8t: Shouldn't we add a space between arguments commas

PR #225, opened 2019-08-09
8.x 3.x config corey

coreylafferty:


source/_patterns/00-config/config.design-tokens.yml
```diff
@@ -41,6 +25,20 @@ gesso:
       gray-5: '#5b616b'
       gray-6: '#323a45'
       black: '#212121'
+    other:
+      yellow:
+        base: '#fad980'
+        light: '#fff1d2'
+        neon: '#ff0'
```
sbrack8t: This seems outside the naming conventions. Should this just be another color key `other.neon-yellow.base`?

source/_patterns/03-layouts/regions/breadcrumb/breadcrumb.twig
```diff
@@ -1,15 +1,17 @@
...
-      {% include '@components/breadcrumb/breadcrumb.twig' with {
-        'breadcrumb': [
-          { 'url': '#', 'text': 'Home' },
-          { 'url': '#', 'text': 'About' },
-          { 'text': 'Current Page' }
-        ]
-      } %}
-    {% endblock %}
-  {# </div> #}
+  {% if not unconstrained %}
```
sbrack8t: This seems confusing checking for the double negative. Can we just check for if `constrain` or if `has-constrain`

source/_patterns/00-config/config.design-tokens.yml
coreylafferty: I'm good with that.  I did update the color.twig to handle things other than just base/dark/light just in case someone uses a different name, but that can just exist as a fallback.

source/_patterns/03-layouts/regions/breadcrumb/breadcrumb.twig
coreylafferty: I did that b/c I wanted the default to have a constrain without having to specify 'constrain: true'... but I see what you mean.  I'll make a change to remove the double negative.

source/_patterns/03-layouts/regions/content/content.twig
```diff
@@ -1,7 +1,11 @@
 <div class="l-content">
```
sbrack8t: In the region section, there's a lot of repetitive code. Thoughts on creating and region template that individual region templates embed
```{# region.twig #}
<section class="l-region l-{{ region_name }}">
  {% if has_constrain %}
    <div class="{{ constrain_classes ?: 'l-constrain' }}">
    {% endif %}
      {% block region_content %}
      {% endblock region_content %}
    {% if has_constrain %}
    </div>
  {% endif %}
</section>

{% embed "@layouts/regions/region.twig"  with {
  region_name: 'content',
  has_constrain: false
}%}
    {% block region_content %}
      
    {% endblock region_content %}
{% endembed  %}
```

coreylafferty: @sbrack8t I went ahead and moved the dummy block content for regions that's just for PL (e.g., the include to the main menu in the navigation region) to the new page-default.twig file b/c that seems more logical with the new structure.

For now I've left the individual region files as I think we should discuss with @dcmouyard.  I'm a bit torn -- yes, we could easily have just a single region.twig file (and a single region.html.twig file for Drupal) given how similar the regions are out of the box.  The header/footer do use different tags than the rest, but we could either incorporate that as an option or just have them all use the same tag.

On the other hand, one of the first things we'd expect someone theming a custom design to do would be to customize the regions, adding new markup and changing/removing constrains, particularly to the header/footer.  They could certainly create their own custom header/footer Drupal and/or PL templates while doing that -- but given that there's a high chance of needing those files, I think we just went ahead and had files for each region so they're already there and ready to go.  If for some reason you didn't need them, it doesn't hurt having them... other than some redundant code.  That said, with the new constrain option it does lend itself better now to a single file approach -- so I'm kind of on the fence.   

A middle ground option would be to just have custom files for header, footer, and maybe content (the most likely to be customized) -- and have the rest use a basic region file.

coreylafferty: Per discussion, I've implemented the middle ground option.

sbrack8t (approved):

Issue #184, opened 2019-01-29
drush gesso --help returns undefined command


ahimsauzi:
Thanks for a great theme, love pattern lab.

When attempting to use drush to create a new theme from gesso, drush gesso --help returns Command "gesso" is not defined. This happens before and after enabling the theme and from any directory (root, web, themes, contrib or gesso). I am using a composer managed drupal 8.6.7 with drush 9.6.0

ahimsauzi:
Looks like example.drush.inc is no longer supported in Drush 9.beta5+.

wats958:
+1. Drush 9 makes the initial theme generation impossible.

I notice that this issue is handled in Gesso 3.x with the included gesso_helper module. It would be nice for support purposes to put this in 2.x as well.

kmonahan:
This fix is in 8.x-3.x.

PR #228, opened 2019-09-03
Bring back `drush gesso` for Drush 9

kmonahan:
As also noted in #184, the `drush gesso` command doesn't work with drush 9. Since drush 9 is what's installed with the Drupal web starter, refactored the drush command to be drush-9-compatible. I also added a couple of notes to the command-line output about things that do _not_ automatically get renamed (namely the gesso_helper module and the gulp tasks) that the user might want to manually update, depending on their use case and how important it is to them that the word "gesso" not appear anywhere.

Most of the code here is not actually new but is taken from gesso.drush.inc with the deprecated methods replaced by whatever the drush 9 equivalent seemed to be.


dcmouyard (approved):
The drush commands work for me.

I added some additional documentation to the readme and fixed some minor issues with coding standards.

kmonahan:
Updated the PR to be merged into 3.1-RC. Will fix those merge conflicts shortly. Edits look good to me. Thanks @dcmouyard !

dcmouyard:
I just noticed that the `gesso-drush.inc` file still gets copied over.

dcmouyard:
If we need to support Drush 8, then we need to keep the `includes/gesso-drush.inc` file and remove it during the copy process. Otherwise we can just delete it if we only need to support Drush 9.

dcmouyard:
Looks like we don’t need to worry about Drush 8 support, so let’s remove the `includes/gesso-drush.inc` file.

kmonahan:
Agreed.

PR #229, opened 2019-09-05
Rollup ==> Webpack

kmonahan:
For discussion later this week, here's what I'm thinking for a setup that uses webpack instead of rollup for bundling.

The main advantage I see is with the split chunk optimization for webpack. The way I've got it set up here (with some extraneous imports for demo purposes that need to be removed before creating a PR for real), code that's shared between multiple entry point files gets put in common.min.js rather than being included multiple times. What I usually do on Drupal sites with more than a trivial amount of JS functionality is create a top-level JS file + library per template and attach the library just to the template(s) I want it in (usually just by sticking attach_library() in the page or node Twig file), and let webpack figure out what's common across pages and what's unique, so that I'm limiting the amount of unrelated JS that gets loaded on a page. I think a similar approach would work with WordPress, since you can use conditionals in whatever function is enqueuing scripts and declare all of other JS files as dependent on the common file.

There's also a minor advantage of being able to switch webpack between development and production modes, so you can have minified, optimized files for production and somewhat easier to debug files for development, although you could probably rig something similar with gulp and rollup.

I don't find webpack harder to configure than rollup, personally. The main disadvantage I see is that webpack + gulp feels a little redundant (although it's typically what I end up doing anyway, because I prefer to handle CSS processing + image processing with gulp) and I have a vague recollection that rollup's tree-shaking capabilities are better than webpack's, although I've yet to be on a project where that's been a key factor.


sbrack8t:
I like it. I was under the false impression that using webpack limited us to a single entry file and 
 restricted the ability to use libraries.

PR #234, opened 2019-09-11
added docs for design token functions to README

coreylafferty:


sbrack8t:
In the example should we include what the output might look like?

coreylafferty:
@sbrack8t you mean like:

example:
```
color: gesso-grayscale(gray-1);
```
which, when compiled to CSS would become:
```
color: #f7f7f7;
```

I think that might be overkill and make the docs harder to read, but I don't feel too strongly either way.

PR #237, opened 2019-09-13
Add ESLint

kmonahan:
Adds ESLint and cleans up any linting errors initially flagged. This involved some light refactoring of bits of the menu functionality, so let me know if you notice anything that broke. The linting rules mostly come from the generator-web-starter (https://github.com/forumone/generator-web-starter/pull/160), in the interest of keeping code style rules consistent-ish across platforms.


sbrack8t (approved):

sbrack8t:
@kmonahan  I noticed you updated the WP version to ES6 class. Did you want to do that here as well?

PR #241, opened 2019-09-20
8.x-3.1-RC renamed font stacks

dcmouyard:
Use name of the font stack rather than the first font-family in the stack.


kmonahan (approved):
👍 This makes good sense.

PR #242, opened 2019-09-20
8.x-3.1-RC Updated sample content with correct punctuation characters

dcmouyard:


kmonahan (approved):
I'm not seeing where this sample content is used in Pattern Lab, since most of the patterns I'm seeing have their own YAML files with content, but changes made here all look good to me.

PR #247, opened 2019-09-24
8.x 3.x c mobile menu

sbrack8t:


source/_patterns/04-components/menus/menu--main/_menu--main.scss
```diff
@@ -3,4 +3,94 @@
...
+    }
+  }
+
+  > .menu__item {
+    display: inline-block;
+    margin-right: rem(gesso-spacing(lg));
+    position: relative;
+
+    > .menu__link {
+      @include link(gesso-grayscale(gray-4), gesso-palette(blue, base));
```
coreylafferty: 'gesso-palette' is no longer a function (you likely need to merge 8.x-3.1-RC into this branch).  In any case, we shouldn't use any palette (i.e., brand or grayscale) colors in the default components.  Instead use colors from the 'colors' section of the design tokens...  maybe something from UI section or create a new one for menus?

source/_patterns/04-components/menus/menu--mobile/_menu--mobile.scss
```diff
@@ -14,6 +14,7 @@ $menu-mobile-submenu-hover-color: #000 !default;
   .menu__item {
```
coreylafferty: Colors defined at the top of this file and throughout it  should be changed to follow other components, using colors from the 'colors' section of the design tokens file.

source/_patterns/04-components/mobile-menu/_mobile-menu.scss
```diff
@@ -3,22 +3,18 @@
 
 $mobile-menu-fallback-bg-color: #f3f3f3 !default;
```
coreylafferty: Colors defined at the top of this file and throughout it  should be changed to follow other components, using colors from the 'colors' section of the design tokens file.

source/_patterns/config/_gesso-theme.scss
```diff
@@ -0,0 +1,238 @@
+$gesso: (
```
coreylafferty: What is this file?

sbrack8t: Artifact from before you restructured the file organization. Removing

.eslintrc.js
```diff
@@ -0,0 +1,50 @@
...
+  "env": {
+    "es6": true,
+    "node": true,
+    "browser": true
+  },
+  "plugins": [
+    "implicit-dependencies"
+  ],
+  "extends": [
+    "eslint:recommended",
```
dcmouyard, line 16: Could we switch from the default eslint rules to Airbnb?

dcmouyard: Hmm, we would need to further refactor a bunch of stuff to make everything pass the airbnb standards. I think we should create a new ticket if we want to tackle that later.

We’ll need to update the current rules, however, to prefer single quotes in JS.

sbrack8t: This is our current rule set for webstarter js projects. We should probably update to match. https://github.com/forumone/generator-web-starter/blob/master/.eslintrc.js

sbrack8t: This also incorporates typescript. Which I dont think we need to do yet.

dcmouyard: My single quote prettier errors were due to me not having `.prettierrc` in the Dockerfile.

dcmouyard:
The mobile menu script works for me! I don’t like how the subnav arrow button prevents users from clicking the parent menu item, so I created a PR for that: #250 

The primary menu script didn’t work for me. I added `<script src="../../../js/dist/primary-menu.min.js?{{ cacheBuster }}"></script>` to the Pattern Lab foot template and added a submenu to one of the main menu items, but it didn’t seem to trigger the script.

sbrack8t:
So you'll need to include the common js file as well as the `primary-menu.min.js`. I'm running into another issue where updates to the header and footer are not making the corresponding changes in the rendered html.

sbrack8t:
Problem solved. Had to fully delete the Pattern Lab folder

dcmouyard:
I’m not sure what changed recently with the mobile menu JS, but I can no longer tab through the menu items. I suspect something changed with how the tabindex is set.

dcmouyard:
I can’t reproduce the mobile menu tabbing error in my D8 test environment. The error originally showed up when I tried to pull these changes into another project.

Turns out I inadvertently changed an aria role name that screwed this up. This isn’t an issue with this PR.

js/src/modules/_Menu.es6.js
```diff
@@ -0,0 +1,258 @@
...
+  open() {
+    // Get position and bounding rectangle of controller object's DOM node
+    const rect = this.controller.getBoundaries();
+    // Set CSS properties
+    if (!this.controller.getIsMenubarItem()) {
+      this.domNode.parentNode.style.position = 'relative';
+      this.domNode.style.display = 'block';
+      this.domNode.style.position = 'absolute';
+      this.domNode.style.left = `${rect.width}px`;
+      this.domNode.style.zIndex = '100';
```
dcmouyard, line 219: Why not use a z-index of 1?

js/src/modules/_Menu.es6.js
```diff
@@ -0,0 +1,258 @@
...
+    } else {
+      this.domNode.style.display = 'block';
+      this.domNode.style.position = 'absolute';
+      this.domNode.style.top = `${rect.height - 1}px`;
```
dcmouyard, line 223: Why subtract a pixel here? Don’t we want it to line up exactly?

js/src/modules/_Menu.es6.js
sbrack8t: To make sure it stays above other items that might have some z-indexing

js/src/modules/_Menu.es6.js
sbrack8t: Trying to avoid any gaps that might cause an inadvertent mouse out.

js/src/modules/_Menu.es6.js
dcmouyard: Since the parent `<li>` has `position: relative`, the only thing that might have a higher z-index that affects this is the `<a>` sibling. I vote to change it to 1 to keep things simple.

js/src/modules/_Menu.es6.js
dcmouyard: Are there any cases where a pixel-rounding issue in a browser might affect this? If not, I’d go with not subtracting 1 pixel. I removed this on a current project since it looked obvious that it didn’t line up.

dcmouyard:
I’ve noticed a bug with the mobile menu when using mobile phones:

- Android v6-10 on Chrome and Firefox
- iOS v10-12 on Chrome and Safari

When the mobile menu is open, scrolling down will close the menu.

This isn’t that noticeable if there aren’t many menu items and scrolling on the mobile menu isn’t needed, but if there are a bunch of items, you can’t scroll down to see them all.

dcmouyard:
Another bug with the mobile menu when using Android phones:

- Android v6-10 on Chrome and Firefox

If you add a search box to the page and add a selector to it in `_MobileMenu.es6.js` then it will be added to the mobile menu. When the mobile menu is open, clicking the text input to enter search terms will close the mobile menu.

dcmouyard:
I just noticed that the primary menu and mobile menu scripts don’t work in IE11.

PR #251, opened 2019-10-04
Allow raw Sass code in design tokens

gustavderdrache:
# Discussion

To better integrate with other design systems, Gesso should have the ability to defer to another system's tokens. As an example, [the USWDS](https://designsystem.digital.gov/) uses `color('<name>')` to access its color tokens. In order to avoid duplication of work, it's better to treat that function call as the authoritative value than to copy and paste it and diverge from the upstream source of truth.

This PR achieves that goal by introducing a custom YAML tag that can be used in `config.design-tokens.yml`. This tag is `!sass`, and indicates to the processor that this value is Sass code to be injected directly and should bypass the normal behavior - this includes token lookup. This means that we can define tokens in terms of other systems' tokens, as well as use lookups when we need to switch sources of truth — that is, we can define a brand blue as coming from another system, or indicate that the primary text color is from another system (see examples below). The code that outputs the `$gesso` Sass map knows to inject the raw Sass directly. This is how we stay in sync with the upstream tokens: when compiled, the Sass code will always use the code instead of a hard-coded value.

# Input Syntax

There are two input syntaxes for this tag: fallback notation and shorthand.

The fallback syntax is a tagged YAML object with `sass` and `fallback` fields. The suggested syntax is to use the block literals, to make it easier to see when the normal token system is being bypassed:

```yaml
example:
  red: !sass
    sass: 'mix(#fff, #f00, 0.5)'
    fallback: '#f00'
```

The second syntax is shorthand notation. This notation is used when there's no need for a separately-labeled fallback. It looks like this:

```yaml
example:
  blue: !sass 'mix(#fff, #00f, 0.5)'
```

This syntax is primarily encouraged for use in token lookups (e.g., `!sass color('red-40')` instead of `red.light`, since the absence of a fallback limits its use as a brand token.

# Output

For the rest of this section, we'll assume these hypothetical tokens:

```yaml
gesso:
  palette:
    brand:
      blue:
        base: !sass mix(#fff,#00f,0.5)
        light: !sass
          sass: mix(#fff,#00f,0.25)
          fallback: '#00f'
```

When this code outputs Sass, the fallback is ignored in both the fallback and shorthand notation:

```sass
$gesso: (
  palette: (
    brand: (
      blue: (
        base: mix(#fff,#00f,0.5),
        light: mix(#fff,#00f,0.25),
        // [snip]
```

However, the code generates an object with `sass` and `fallback` values in both cases. Templates (such as Pattern Lab swatches) that expect simple tokens can check for `token is iterable` in Twig and read the array values. Notice that the shorthand syntax still output an array - this way, the template code always knows when the normal tokens were bypassed and can indicate to anyone reviewing PL that the source of truth for the token isn't stored directly in the design tokens file.

```yaml
gesso:
  palette:
    brand:
      blue:
        base:
          &a1
          sass: mix(#fff,#00f,0.5)
          fallback: null
        light:
          &a2
          sass: mix(#fff,#00f,0.25)
          fallback: "#00f"
```

(Aside: for some reason, the YAML library likes to output anchors for these values. I don't know why, and it's very annoying. However, it doesn't break the Pattern Lab build, so I'm going to ignore it for now.)

# Example: Swatches

In addition to the above code changes, I modified `_color-item.twig` to demonstrate this behavior. Using the example above, the swatches generated for blue.base and blue.light are below. Note that the base color appears broken because there's no fallback - the generated css is `background-color: mix(#fff, #00f, 0.5)`, which doesn't work in Firefox. The second swatch, since it has a fallback, is able to display an approximation of the compiled Sass value (and we can see it when the token is referenced via `gesso-color()` or `gesso-brand()`).

image

When used as token lookup, the rendered Sass is used directly. The default Gesso text links use `brand.blue.base`, and we can see that the compiled styles correctly use the value of `mix(#fff, #00f, 0.5)` (`#0101ff`, according to SassMeister):

```css
a {
  background-color: transparent;
  color: #0101ff;
  outline-offset: 0.125em;
  /* [snip] */
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,133 +97,146 @@ gulp build
 
 To access the Pattern Lab instance, append `/pattern-lab/index.html` to your site URL and theme directory (e.g. http://localhost:8080/themes/gesso/pattern-lab/index.html) or, if developing locally, just open that index.html file directly in the browser from your file system.
 
-
 ### Design Tokens
-Gesso uses a configuration file 'source/_patterns/00-config/config.design-tokens.yml' to manage the theme's design tokens and automatically generate the global sass map for styling and patterns to represent the theme's design tokens. The default gulp command will monitor changes in the config and rebuild all neccessary assets. To rebuild the theme assets a single time run `gulp gessoBuild`
+
+Gesso uses a configuration file 'source/\_patterns/00-config/config.design-tokens.yml' to manage the theme's design tokens and automatically generate the global sass map for styling and patterns to represent the theme's design tokens. The default gulp command will monitor changes in the config and rebuild all neccessary assets. To rebuild the theme assets a single time run `gulp gessoBuild`
+
+#### Design Token Colors
+
+In addition to using hex values for colors, you can also use Sass functions and
+provide a fallback hex value to be used in Pattern Lab. This is useful when
+pulling in an external design system, such as [USWDS](https://designsystem.digital.gov/).
+
+For example:
+
+```
+gesso:
+  palette:
+    brand:
+      blue:
+        base: '#0071bc'
+        light: !sass
+          sass: 'lighten(#0071bc, 20%)'
+          fallback: '#23a7ff'
+```
 
 #### Design Token Functions
-The following Sass functions can be used to access the tokens defined in `config.design-tokens.yml`.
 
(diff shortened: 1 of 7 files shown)


dcmouyard (approved):
Works like a charm Alex!

I added documentation for this in the README.

PR #253, opened 2019-10-04
Change Focus handlers

sbrack8t:


js/src/primary-menu.es6.js
```diff
@@ -8,7 +8,7 @@ import Drupal from 'drupal';
 
   Drupal.behaviors.primaryNav = {
     attach(context) {
-      const submenuNode = context.querySelector('.menu--main');
+      const submenuNode = context.querySelector('.menu--submenu');
```
kmonahan: I don't think this is correct. The MenuBar should be the horizontal menu with the top-level links, which is menu--main. (Why the variable is called submenuNode is an excellent question. I think we should change that to something more sensible like menuNode.) I'm basing this off of https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html, where the top-level menu has the role of 'menubar'.

js/src/modules/_Menu.es6.js
```diff
@@ -175,7 +183,8 @@ export class PopupMenu extends Menu {
   }
 
   setFocusToItem(newItem) {
-    newItem.focus();
+    //eslint-disable-next-line
```
kmonahan: Can we remove this? What linting problem are you hitting with the next line?

kmonahan (changes requested):
See comment about changing the menu selector. Happy to discuss that one further.

js/src/primary-menu.es6.js
sbrack8t: That makes sense. I'll update.

js/src/modules/_Menu.es6.js
sbrack8t: I had a debugger statement after the line. Forgot to remove the comment

PR #258, opened 2019-10-14
moved label class logic from D8 to PL template

coreylafferty:


dcmouyard (approved):
I fixed a missing comma. Everything else looks good. 👍

PR #262, opened 2019-10-22
updated PL version, renamed component folders to be singular, fixed m…

coreylafferty:
…essages class, fixed button class order


dcmouyard (approved):
I added a commit to fix missing EOL characters. Everything works like a charm!

PR #264, opened 2019-10-23
8.x 3.x test search box mobile

sbrack8t:
@dcmouyard 
I added a check in the resize handler to verify that screen width actually changed. The input and scroll issues appear resolved in browserstack on ios and android devices and  Apparently in mobile browsers using the text input or overscrolling triggers height/resize events.  There's an over-scroll issue in the ios safari(browserstack) where you can see the content underneath the menu if you over-scroll. But was unable to replicate on the actual device. I would also recommend that we put the search bar at the top as android/google device can sometimes overlay the keyboard versus triggering a resize.

### DIFF
--- a/js/src/modules/_MobileMenu.es6.js
+++ b/js/src/modules/_MobileMenu.es6.js
@@ -294,12 +294,18 @@ class _MobileMenu {
     this.close();
 
     let resizeTimeout = false;
+    let lastWindowWidth = window.innerWidth;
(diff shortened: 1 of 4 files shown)


dcmouyard:
@sbrack8t The fix works for me!

I recommend creating a new PR for this fix without the search bar getting added to the page or mobile menu, since that won’t be on every project.

PR #271, opened 2019-11-12
Consolidate and add to unit conversion functions

kmonahan:
This came out of the work on gesso-uswds, as many of the USWDS design system tokens are in units other than pixels (usually rems). Depending on the current design token settings, a call to `rem(gesso-spacing(xs))` might pass a pixel value to the function or might pass a rem value. (Or, since users can presumably put any valid size in the design token YAML file, this could also be an em value.)

The following will now work:
```
$base-font-size: 16px; // Base font size MUST be in pixels or the functions will throw an error.
rem(16px, $base-font-size); // 1rem
rem(1em, $base-font-size); // 1em
rem(px(1em, 12px), $base-font-size); // 0.75rem
rem(1rem, $base-font-size); // 1rem
rem(16, $base-font-size); // 1rem. Unitless values are assumed to be pixels.
```
Same for the em() function. (As before, the second parameter is optional and will default to the Gesso font base size if not specified.)

As noted in the comments on the Sass file, em to rem conversion is imperfect, since the em base size might be different than the rem base size. If that's the case, you're better off following the third example above, where you convert the em to pixels and the resulting size to rems.

Unitless handling is in there as a convenience because some other px-to-rem conversion functions I've seem (namely Foundation's `rem-calc()`, which I used to use a bunch) use unitless values. I'd still recommend we explicitly specify the unit when setting any values within Gesso itself.

### DIFF
--- a/source/_patterns/00-config/00-functions/_functions.em.scss
+++ b/source/_patterns/00-config/00-functions/_functions.em.scss
@@ -1,7 +1,0 @@
-// @file
-// Em conversion
-
-// Convert pixel values to em.
-@function em($px, $base: gesso-get-map(typography, base-font-size)) {
-  @return ($px / $base) * 1em;
-}
--- a/source/_patterns/00-config/00-functions/_functions.rem.scss
+++ b/source/_patterns/00-config/00-functions/_functions.rem.scss
@@ -1,7 +1,0 @@
-// @file
-// Rem conversion
-
-// Convert pixel values to rem.
-@function rem($px, $base: gesso-get-map(typography, base-font-size)) {
-  @return ($px / $base) * 1rem;
-}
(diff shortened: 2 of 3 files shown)


sbrack8t:
This looks good to me.

dcmouyard:
I think the `rem()` and `em()` functions should error if they don’t return the correct unit.

kmonahan:
@dcmouyard, what use case are you thinking of?

dcmouyard:
As an example, if I set `font-size: rem($something);` I expect the font-size to be in rems, no other units.

This is especially critical in cases where I’m doing calculations. (e.g., `line-height: rem($thing1) / rem($thing2);`

kmonahan:
@dcmouyard, what I meant was, what would you be passing in to the line-height value in that calculation where you're getting something other than rems back?

dcmouyard:
My apologies, I was basing my comments on your example of `rem(1em, $base-font-size); // 1em`. 

The code in this PR makes sure that the units from the `px()`, `em()`, and `rem()` functions are always correct.

kmonahan:
Ha, sorry! Didn't even notice that -- total copy/paste error on my part. It would yield 1rem.

PR #272, opened 2019-11-12
Add create_attribute function for Pattern Lab

kmonahan:
Pulls in a `create_attribute` function so it can be used in PL templates. The PL version isn't exactly equivalent to the Drupal version (you can't chain functions), but lets you do something like this in your PL template:

```
{% set link_attributes = link_attributes|merge({
  "class": link_attributes["class"]|merge([
    'usa-accordion__button',
    'usa-nav__link'
  ]),
  "aria-expanded": "false"
}) %}
<button {{ create_attribute(link_attributes) }}><span>{{ item.title }}</span></button>
```

### DIFF
--- a/pattern-lab-config.json
+++ b/pattern-lab-config.json
@@ -239,6 +239,10 @@
         {
           "file": "source/_twig-components/functions/url.function.php",
           "functions": ["addUrlFunction"]
+        },
+        {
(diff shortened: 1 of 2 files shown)


dcmouyard (approved):
Works like a charm!

PR #289, opened 2019-11-20
Update skiplinks

dcmouyard:
This work moves the skiplinks to its own folder in components and updates the styles to use design tokens.

I debated pulling the skiplinks markup into that component, removing it from `_meta/00-head.twig` and adding it to `_patterns/06-pages/_page-default.twig`. Should I do that as well?


kmonahan:
> I debated pulling the skiplinks markup into that component, removing it from `_meta/00-head.twig` and adding it to `_patterns/06-pages/_page-default.twig`. Should I do that as well?

Yes, I think so. That feels more consistent with how we're treating other components. Also thinking you may want to replace the skiplinks markup in https://github.com/forumone/gesso/blob/8.x-3.x/templates/layout/html.html.twig with the included component.

kmonahan (approved):
Looks good! @coreylafferty, this is ready to merge.

sbrack8t (approved):
Ship it!

PR #290, opened 2019-11-20
Add ability to show demo button examples in Pattern Lab

dcmouyard:
This work adds examples of button, link, and disabled versions of buttons in Pattern Lab.

image


kmonahan (approved):
Nice! I like this idea.

@coreylafferty, this one's ready to merge.

PR #291, opened 2019-11-20
Add icon component that uses an SVG sprite

dcmouyard:
This work adds an SVG sprite (with individual source SVGs located in the `images/_icons` directory) and an icon component that uses that sprite.


images/icons.svg
```diff
@@ -0,0 +1,6 @@
...
+</svg>
```
kmonahan, line 6: What was the thought behind manually creating the sprite versus using something like https://www.npmjs.com/package/gulp-svg-sprite? (I will admit that I'm biased toward "getting robots to do my work for me.")

dcmouyard: I decided to do it manually at first just to see whether this approach is worth putting into Gesso. I’m definitely in favor of creating a gulp task to compile the SVG sprite, whether that’s done in this PR or a follow-up one.

dcmouyard: By the way, I’ve used the gulp-svg-sprite plugin on a previous project and it definitely works for this approach.

kmonahan: Fair enough. I agree that creating a gulp task would be a nice enhancement, but I don't think lack of one needs to block merging this. It can be a future PR.

kmonahan (approved):
@coreylafferty, this one's ready to be merged.

PR #293, opened 2019-11-21
 Update list components

dcmouyard:
**Changes:**
* Updates list components to make it possible to directly pull variants into other templates
* Change list items to use content instead of text
* Update list mixins to match coding standards


kmonahan (approved):
lgtm! @coreylafferty, ready to merge.

PR #295, opened 2019-11-21
Add gutter width design token

dcmouyard:
**Changes**
* Add gutter-width design token
* Update grid mixins with gutter-width token
* Update constrain mixins with gutter-width token


kmonahan (approved):
Nice! I think this will be useful. @coreylafferty, another one ready to merge.

PR #299, opened 2019-11-26
Add dark background option for Pattern Lab components

dcmouyard:
Changes
* Add a dark background and adjust padding for PL components that need one such as `--inverse` variants

This will be applied by default to any component that has `--inverse` in the component name.

**Example from a recent project:**

Before
image

After
image

### DIFF
--- a/source/pattern-lab.scss
+++ b/source/pattern-lab.scss
@@ -333,3 +333,25 @@ $swatch-padding: 1rem;
 .pattern-lab-text-style__label {
   text-transform: uppercase;
 }
+
+// For components used on a dark background. (e.g., `--inverse` variants)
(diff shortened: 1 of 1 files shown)


kmonahan (approved):
Nothing that needs to block merge of the initial version, but I think it would be nice here to have a little more flexibility in triggering the dark background. I could imagine having some component that will only ever be displayed on a dark background, so creating an inverse variant wouldn't make sense.

dcmouyard:
For those cases, I would add the ID selector for that component to this section of `pattern-lab.scss`.

I got tired of copying this snippet from project to project (where I just used the specific component selector) and wanted to add a default option.

PR #278, opened 2019-11-18
Update documentation for patterns

dcmouyard:
On projects where I use states on patterns to track their progress, I had to add a bunch of files just to add this info. This PR adds those missing markdown files for those components.

This PR also updates all pattern documentation to match our coding standards.

PR #300, opened 2019-11-26
Lint Pattern Lab Markdown Files

kmonahan:
In https://github.com/forumone/gesso/pull/278, Dan updated the Pattern Lab documentation to include markdown files for patterns that were missing them and to standardize formatting. To maintain the coding standards now in place, I wrote some linting scripts for the Pattern Lab markdown files.

There's room to build on it further in the future, but initially it'll display errors if:

- You have a Twig file that does not begin with an underscore but no corresponding markdown file (looking specifically two or more levels down, so it'll ignore the templates and pages directories.
- You have a markdown file that doesn't have YAML front-matter.
- You have a markdown file that specifically doesn't have the title defined in front-matter. (I'm not linting for el, since that doesn't seem to be used for all patterns)
- You have a Variables heading that doesn't use the format `__Variables:__`
- You have a list of variables that doesn't strictly follow the format `* variable_name: [type] Description`.

This PR also fixes a couple remaining places where those rules weren't in use.


dcmouyard (approved):
Looks good to me! 👍

PR #302, opened 2019-12-03
added modifier classes to view componenent, made other files consiste…

coreylafferty:
…nt in how mod classes are added

### DIFF
--- a/source/_patterns/03-layouts/breadcrumb/breadcrumb.twig
+++ b/source/_patterns/03-layouts/breadcrumb/breadcrumb.twig
@@ -1,6 +1,11 @@
+{% set constrain_classes = [
+  'l-constrain',
+  constrain_modifier_classes ? constrain_modifier_classes,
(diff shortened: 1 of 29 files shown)


dcmouyard (changes requested):
I really like these changes! I think we should also adjust variants to follow this pattern as well.

Example:
```
{% set button_modifier_classes -%}
  button--small {{ modifier_classes }}
{%- endset %}

{% include '@components/button/button.twig' with {
  'url': url,
  'text': text,
  'modifier_classes': button_modifier_classes,
} %}
```

Should be updated to:
```
{% set classes = [
  'button--small',
]|join(' ')|trim %}

{% include '@components/button/button.twig' with {
  'url': url,
  'text': text,
  'modifier_classes': classes,
} %}

```

dcmouyard:
Hmm, I don’t think my requested changes should block the merge.

dcmouyard (approved):

PR #304, opened 2019-12-11
Color token order

dcmouyard:
**Changes:**
- Refactor color twig template to output colors in the same order as they are defined in the design tokens yaml file
- Change color headings from `<h4>` to `<h3>` to match page heading structure

This work doesn’t change the current order colors are shown in Pattern Lab, but it allows one to change the order in the design tokens yaml file and see that change reflected in the generated Pattern Lab page.


kmonahan (approved):
Nice! Tried rearranging the order locally, and it's working as expected.
Screen Shot 2019-12-12 at 9 08 54 AM
Screen Shot 2019-12-12 at 9 09 03 AM

PR #307, opened 2019-12-11
Update Pattern Lab to 5.4.0

dcmouyard:


kmonahan (approved):
Looks good to me! Did see the message "Could not find uikit with name uikit-polyfills defined within patternlab-config.json, or it is not enabled." when building, but based on https://github.com/pattern-lab/patternlab-node/pull/1098, that seems to be a known issue with PL (some kind of name collision) and wasn't causing any errors I could find.

PR #309, opened 2019-12-23
Add node version specifications

kmonahan:
Makes it clearer to users what version of Node you'll need for Gesso:

- Adds Node.js to the global prerequisites
- Adds a `.nvmrc` file to use 12.x (current LTS)
- Sets 10.x as the minimum node version allowed in package.json and enables engine-strict, so the user will get an error if running `npm install` on an unsupported version. (12.x is the recommended version, but 10.x is the minimum version needed for fs.promises support.)


dcmouyard (approved):
Looks good to me. 👍

sbrack8t (approved):
💯

PR #311, opened 2020-01-14
Update vertical spacing

### LINKED ISSUE #306: Vertical spacing
Currently in Gesso, we remove the top margins on most things and only set a bottom margin, usually based on one of the spacing design tokens.
```
p {
  margin: 0 0 rem(gesso-spacing(sm));
}
```

One notable exception to this is used for headings, where we add top margin if there’s a sibling just before the heading.
```
%hN {
  margin: 0 0 rem(gesso-spacing(sm));

  // Add top margin when headers are preceded by other elements (p, ul, etc.)
  * + & {
    margin-top: 1em;
  }
}
```

This approach means that most things have a bottom margin, and in certain cases we want to remove the bottom margin on the last item in something.
```
.message__content {
  > *:last-child {
    margin-bottom: 0;
  }
}
```

The USWDS uses an approach where items don’t get top or bottom margins by default, but get added only if items are before or after it, similar to how we’re currently handling headings.
```
p {
  margin: 0;

  // Add top margin when preceded by another element.
  * + & {
    margin-top: 1em;
  }

  // Add top margin to item directly after this.
  + * {
    margin-top: 1em;
  }
}
```

How do folks feel about switching to the USWDS approach?

dcmouyard:
Fixes #306 

**Changes**
* Add new `vertical-spacing` mixin
* Update top & bottom margin styles to use this new mixin where applicable

**Notes**
One thing to note about this approach: elements won’t have top/bottom margins if there are wrappers around individual items.

This will have the appropriate vertical spacing between items:
```
<p>This is a paragraph.</p>
<p>This is another paragraph.</p>
```

This will **not** have the appropriate vertical spacing between items:
```
<div>
  <p>This is a paragraph.</p>
</div>
<div>
  <p>This is another paragraph.</p>
</div>
```


sbrack8t:
This approach seems a little heavy-handed particularly with the inclusion of the wild card selector. What's the benefit to this mixin versus applying margin top/bottom directly to an element

dcmouyard:
@sbrack8t The rationale behind switching to this technique can be found here: https://github.com/forumone/gesso/issues/306

sbrack8t:
@dcmouyard Thanks for the context

coreylafferty:
@dcmouyard For most basic elements, this looks good.  However for a few of our example components this doesn't work well and breaks the prior layout a bit.  At first glance:

- There is not sufficient space between the Site Name and the Main Menu on example pages
- The breadcrumb items are no longer aligned in the middle of the breadcrumb region's background color
- You removed some of the styles for the card__footer that ensured that the Tags and Read More links are always at the bottom of the card (i.e., they appear in the same part of the card regardless of the card content's height).

Can you address these and I'll take another review pass?

dcmouyard:
@coreylafferty I addressed those three specific cases. Are there any others?

coreylafferty:
Thanks, @dcmouyard.  I did a more thorough pass and here's what I've found:

- Button: Since this element in inline, when there is more than one button next to each other the following ones are misaligned (pushed down)
- Dropbutton: The clickable part is misaligned
- Article: article__footer makes space even if it's empty.  We should probably add an if statement around it so its markup isn't created if it has nothing in it.  Or remove the margin.
- fieldset__legend is too close to form items (see Checkboxes Form Item / Radios Form Item)
- form-items too far apart b/c they have both top and bottom margin (see Checkboxes Form Item / Radios Form Item)
- Pipeline list items misaligned
- Pager:  'Previous' link misaligned (in both mini and normal page)
- Tabs:  first tab misaligned
- Footer region (l-footer) does not get any top margin.  It either needs some or needs some padding so there's space between footer content and main content.
- Mobile menu: third level of mobile menu has too much top margin 
- Page title on our example pages: might need a custom rule for the page-title?  it gets inconsistent space after it depending on what follows (e.g., a <p> or a l-section) when I think typically we'd want consistent space after the page-title.  maybe give it its own bottom margin or move it somewhere else?

dcmouyard:
> Article: article__footer makes space even if it's empty. We should probably add an if statement around it so its markup isn't created if it has nothing in it. Or remove the margin.

There was already an if statement wrapping the article footer to hide it if empty. Something’s screwy with the default content inside that footer block (not immediately clear to me why it wasn’t getting output) so I set the `has_footer` variable to false by default.

Everything else should be fixed now.

coreylafferty (approved):

Issue #313, opened 2020-01-15
Aggregate JavaScript Breaks Site


ghost:
Hi there, not exactly sure if this a Gesso thing or not, but thought I would post it here to see if anyone has this same issue. If I check "Aggregate JavaScript files" in Drupal, a few things on the site break including the admin toolbar. They work fine with that option unchecked. Checking the console, I see the following error:
```
SyntaxError: "" string literal contains an unescaped line break
```

Which if I look into the compiled JS and find the line number (I think this is the one) I see this area:

```
/***/ "./node_modules/core-js/internals/a-function.js":
/*!******************************************************!*\
  !*** ./node_modules/core-js/internals/a-function.js ***!
  \******************************************************/
/*! no static exports found */
/***/ (function(module, exports) {

eval("module.exports = function (it) {\n  if (typeof it != 'function') {\n    throw TypeError(String(it) + ' is not a function');\n  } return it;\n};\n//# sourceURL=[module]\n//# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbIndlYnBhY2s6Ly8vLi9ub2RlX21vZHVsZXMvY29yZS1qcy9pbnRlcm5hbHMvYS1mdW5jdGlvbi5qcz8xYzBiIl0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiJBQUFBO0FBQ0E7QUFDQTtBQUNBLEdBQUc7QUFDSCIsImZpbGUiOiIuL25vZGVfbW9kdWxlcy9jb3JlLWpzL2ludGVybmFscy9hLWZ1bmN0aW9uLmpzLmpzIiwic291cmNlc0NvbnRlbnQiOlsibW9kdWxlLmV4cG9ydHMgPSBmdW5jdGlvbiAoaXQpIHtcbiAgaWYgKHR5cGVvZiBpdCAhPSAnZnVuY3Rpb24nKSB7XG4gICAgdGhyb3cgVHlwZUVycm9yKFN0cmluZyhpdCkgKyAnIGlzIG5vdCBhIGZ1bmN0aW9uJyk7XG4gIH0gcmV0dXJuIGl0O1xufTtcbiJdLCJzb3VyY2VSb290IjoiIn0=\n
/***/ }),
```

The line in question is the eval line specifically. Since it's referencing node-modules, is this part of gesso? Thanks!

ghost:
[...]

PR #316, opened 2020-01-17
adds details polyfill for older browsers and updates details components

coreylafferty:


dcmouyard (commented):
This PR works for me in IE Edge and IE 11 when viewing the details component in Pattern Lab, but not in Drupal.

Strangely, when turning off JS aggregation the items on a node edit page that are usually in details end up in vertical tabs instead.

dcmouyard:
I get the same details -> vertical tabs on node edit pages when JS aggregation is turned off on `8.x-3.x` so it’s not something introduced in this PR.

dcmouyard (approved):
After talking this through with Corey, we think the problem I had was due to #313 , and the un-aggregated JS polyfill worked for me when manually showing the details summary.

kmonahan:
Was just about to ask if this was related to #313. I'm looking into that one now, but in the mean time, testing with `gulp build` rather than `gulp` should get you the proper behavior.

PR #318, opened 2020-01-20
Use dedicated forumone/gesso image

gustavderdrache:
This replaces the image build steps for Gesso with the dedicated  [forumone/gesso](https://hub.docker.com/r/forumone/gesso) image on the Docker Hub. This image is Alpine-based to reduce disk space usage of the base image, and the centralized image should make it easier to ship updates to the build environment (for example, raising the Node.js or PHP versions).

This has been tested with fresh copies of Gesso as well as a few existing projects, so it should be ready for more widespread adoption.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,4 @@
-FROM php:7.3-cli
-
-ENV NODE_VERSION=v12.13.0 \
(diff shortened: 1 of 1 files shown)


dcmouyard (approved):
This change worked for me on two separate projects plus my Gesso/Gus dev project.

PR #319, opened 2020-01-20
Change webpack source map methods

kmonahan:
Addresses #313 . When not in production mode (running `gulp`), switches the webpack source map generator tool from `eval-source-map` to `source-map`. `eval-source-map` is typically recommended for development because it's faster when rebuilding files (or so https://webpack.js.org/configuration/devtool/ told me) but apparently doesn't play nicely when Drupal's aggregate JS and CSS are turned on.

This PR also switches the source map tool to none when in production mode (running `gulp build`), again following the recommendations of the webpack docs. In both cases, these are the default starting options, but developers on individual projects can override them by editing webpack.config.js for their individual projects.

Finally, I fixed some formatting issues that the ESLint plugin for PHPStorm flagged for me while I was working.

*Note for Testing*
If you have an existing testing site, you'll need to run `f1 build` or otherwise rebuild your Gesso container in order to copy the updated webpack config into the container before trying to build the front-end.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -2,12 +2,12 @@ const TerserJsPlugin = require('terser-webpack-plugin');
 const glob = require('glob');
 const path = require('path');
 
-module.exports = (mode) => {
+module.exports = mode => {
   const isProduction = mode === 'production';
   const isDevelopment = !isProduction;
   const sourceFiles = glob.sync('js/src/*.es6.js');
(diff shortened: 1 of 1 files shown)


dcmouyard (approved):
Looks good to me! 👍

dcmouyard:
Fixes #313

PR #320, opened 2020-01-22
update font size and spacing key naming

coreylafferty:


coreylafferty:
Since I've added a new Twig filter, you may need to run `f1 build` if you're using a web-starter instance to test this.

source/_patterns/00-config/config.design-tokens.yml
```diff
@@ -265,11 +265,18 @@ gesso:
...
+    3: 24px
+    4: 32px
+    5: 40px
+    6: 48px
+    7: 56px
+    8: 64px
+    9: 72px
+    10: 80px
+    12: 96px
+    15: 120px
```
dcmouyard: Should we add `1.5: 12px` and `2.5: 20px`? Those are common spacings I’ve seen on projects. I know we’re currently using 12px in fieldsets and 20px in form items.

dcmouyard (approved):
Everything looks good to me! 👍 

The discussion around adding 12px and 20px shouldn’t block this PR.

source/_patterns/00-config/config.design-tokens.yml
kmonahan: +1 for adding 1.5 and 2.5. Not only are those common spacings but including that provides an example of how to handle spacing needs not already included here without resorting to putting rem(20px) all over your project.

sbrack8t: Should the naming only include integers versus floats? If you need an in-between value you can always perform an operation on that value. Using integers would also make it easier to programmatically map over the tokens to create custom classes for example.

dcmouyard: I think the eventual goal is to replace these design tokens with a function (probably same name as current function that grabs these tokens) that will return the px value for some multiple of a base size. In out current setup the base size would be 8px, and the function would return `(N x 8px)`.

I do like the idea of mapping these values to generated CSS classes though... Perhaps that mapping can skip non-integers?

dcmouyard: Or we can define a list of which spacing values to use in that specific partial.

```
$sizes: 0, 1, 2, 3, 4, 5;

@each $size in $sizes {
  .u-spaced-#{$size},
  .u-spaced-#{$size}-above {
    margin-top: #{rem(gesso-spacing($size))} !important;
  }

  .u-spaced-#{$size},
  .u-spaced-#{$size}-below {
    margin-bottom: #{rem(gesso-spacing($size))} !important;
  }
}
```

sbrack8t: The yml token creates a sass map for us.  We can use it to  loop through so you don't have to maintain the sizing variable. I guess the float doesn't really matter other than changing how the float renders as a key for use in the class names to avoid something like `.u-space-1.5`

sbrack8t: I'm +1 on the change.

coreylafferty: added 1.5 and 2.5 key values.

PR #322, opened 2020-01-25
Watch Markdown files; Adjust regex for variables section

### LINKED ISSUE #321: Error: TypeError: Cannot read property '1' of null at lintPatterns
Reported by @tomealter in Slack:

Pattern Lab linter is failing when linting what should be a valid file. File
```
---
el: .callout
title: Callout
---

__Variables:__
* callout_content: [object] Callout out formatted text. 
```

Error seems to be specifically with `__Variables:__` as linter works when that is removed.

kmonahan:
Closes #321 . Adjusts the regex that identifies the Variables section of a pattern lab markdown file to handle the case where there is no new line or carriage return at the end.

Also adds a watch task for markdown files, so you can get immediate feedback from the linter when editing those files without having to also update the twig file.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -164,6 +164,13 @@ const watchFiles = () => {
     { usePolling: true, interval: 1500 },
     bundleScriptsDev
   );
+  watch(
+    [
+      'source/**/*.md',
+    ],
+    { usePolling: true, interval: 1500 },
+    lintPatterns
+  )
 };
 
 const buildStyles = (exports.buildStyles = series(lintStyles, compileStyles));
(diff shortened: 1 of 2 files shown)


dcmouyard (approved):
The changes work for me as advertised. 👍 

Regarding the missing newline at the end of the file that caused this issue: we’ve been following the [Drupal coding standards](https://www.drupal.org/docs/develop/standards/coding-standards#indenting) which requires each file to end with a newline character `/n`. I’ve been relying on my editor settings to enforce this, but having this tested during linting would help (with a helpful error message).

PR #323, opened 2020-01-25
Add unique ID filter to Twig

kmonahan:
Drupal 8 includes [Html::getUniqueId](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Html.php/function/Html%3A%3AgetUniqueId/8.2.x), which ensures that IDs aren't repeated on a page. This PR makes that functionality available as a Twig filter. For Pattern Lab, a unique ID is generated via the uniqid() function.

The typical use case for this is when you need to include an ID in multiple places in a pattern, such as in another element's ARIA attribute (as in the breadcrumbs pattern, which I updated here to use the new filter) or in a data attribute, particularly where the component might be used multiple times on a page. I usually find myself needing this with accordions.

### DIFF
--- a/source/_patterns/04-components/breadcrumb/breadcrumb.twig
+++ b/source/_patterns/04-components/breadcrumb/breadcrumb.twig
@@ -3,9 +3,11 @@
   modifier_classes ? modifier_classes,
 ]|join(' ')|trim %}
 
-<nav aria-labelled="breadcrumb-label" class="{{ classes }}" role="navigation">
+{% set breadcrumb_label = 'breadcrumb-label'|unique_id %}
+
+<nav aria-labelled="{{ breadcrumb_label }}" class="{{ classes }}" role="navigation">
   <div class="l-constrain">
-    <h2 class="breadcrumb__title visually-hidden" id="breadcrumb-label">{{ heading|default('You are here') }}</h2>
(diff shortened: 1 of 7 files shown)


source/_patterns/04-components/breadcrumb/breadcrumb.twig
```diff
@@ -3,9 +3,11 @@
   modifier_classes ? modifier_classes,
 ]|join(' ')|trim %}
 
-<nav aria-labelled="breadcrumb-label" class="{{ classes }}" role="navigation">
+{% set breadcrumb_label = 'breadcrumb-label'|unique_id %}
```
dcmouyard: I think we should rename this variable to `breadcrumb_id` to make its purpose clear.

dcmouyard (changes requested):
Everything looks good to me, except for the ID variable.

dcmouyard:
Should we also add this function to the pager label in `source/_patterns/04-components/pager/pager.twig` and `source/_patterns/04-components/pager/pager--mini/pager--mini.twig`?

kmonahan:
@dcmouyard , changes made.

dcmouyard (approved):
Looks good to me 👍

Issue #325, opened 2020-01-30
Vulnerabilities when running npm install


ebremner17:
When running npm install, using nodejs 13.7.0 or 12.14.6 and gulp 4.0.2, there are 46 vulnerabilities (5 low, 3 moderate and 38 high).  See attached file for the fixes suggest form npm audit.

Here is the error message when running npm install:
found 46 vulnerabilities (5 low, 3 moderate, 38 high) in 32982 scanned packages
  run `npm audit fix` to fix 41 of them.
  1 vulnerability requires semver-major dependency updates.
  4 vulnerabilities require manual review. See the full report for details.

[npm_audit.txt](https://github.com/forumone/gesso/files/4135675/npm_audit.txt)

ghost:
Following. I got the same with the last build I used as well. I'm not as familiar with NPM so I didn't know if I could/should/know how to fix the last 5.

kmonahan:
This was addressed in https://github.com/forumone/gesso/pull/326. As noted in the PR, you'll still see a few vulnerabilities that trace back to @pattern-lab/core, which needs to update some of its dependencies for those to be resolved.

A word of caution if you apply the work in 326 to a site that's already underway: that PR also gets Prettier up and running on scss files again, which means you'll probably start seeing linting errors you never saw before. You can either fix them (many text editors have Prettier integrations that'll do the work for you) or set the `prettier/prettier` rule to false in `.stylelintrc.yml`.

PR #326, opened 2020-01-31
Dependency Updates + Prettier Plugin Swap

kmonahan:
Addresses as much as can be addressed for #325 . There are still 3 remaining low security vulnerabilities. All trace back to pattern-lab core, which is using an old version of the chokidar module, which in turn has outdated dependencies. The Pattern Lab team will need to switch to version 2 or 3 of chokidar in order to resolve those remaining problems. They're out of our control.

This PR does the following:

- Updates all dependencies with security vulnerabilities that could be updated via `npm audit fix`
- Replaces the `prettier-stylelint` package, which is no longer supported, with the `stylelint-prettier` package, which is. These are not exactly identical (there's some discussion of the differences in https://github.com/hugomrdias/prettier-stylelint/issues/34), but stylelint-prettier allows us to use the Prettier rules already used by ESLint linting SCSS.
- Cleans up all of the linting issues found once stylelint-prettier was up and running. (Although prettier-stylelint was in place previously, given the number of formatting issues flagged, I'm not sure whether it was actually running code through Prettier to clean up formatting.)

If you _don't_ want Prettier rules being flagged by stylelint, set the prettier/prettier rule to false in your .stylelintrc.yml file.

*Testing Notes*
If you're using the Docker/web starter generator, you will need to run `f1 build` first.


coreylafferty (approved):

dcmouyard (approved):
👍

PR #327, opened 2020-01-31
Share tokens with js

kmonahan:
Creates a JavaScript file from the design tokens YAML file so that a site's CSS and JS can share a single source of truth. Updates the menu JS to use these tokens, so that the breakpoint and z-index values come from the design tokens instead of being hard-coded.

All design tokens are available for import, so you can use whatever you need, but the place I most often find myself needing to sync values between CSS and JS is with breakpoints. Example usage (also in the README):
```
import { BREAKPOINTS } from '../constants/_GESSO.es6';
if (window.matchMedia(`min-width: ${BREAKPOINTS.desktop}`).matches) {
  // Some script that should only run at the Gesso desktop breakpoint.
}
```

If the token value is a Sass function, the JavaScript will use the fallback value, if available. If there is no fallback value, the token will be omitted from the JavaScript objects. If you want to use some complicated Sass functionality for a token, though, it's probably not a good candidate for exporting to JS. I'm thinking of this more for fixed sizes, breakpoints, or durations, if you need a script to be timed with a CSS transition.


lib/renderJs.js
```diff
@@ -0,0 +1,33 @@
...
+const SassValue = require('./SassValue');
+
+function createJsonMap(data) {
+  return JSON.stringify(
+    data,
+    (key, value) =>
+      value instanceof SassValue
+        ? value.fallback
+          ? value.fallback
+          : undefined
```
gustavderdrache: Accessing a nonexistent property in JS returns `undefined` already - so it seems like this is superfluous, unless you want to collapse falsey values (`''`, `0`, `false`, `null`) to `undefined` as well. Even then, I argue it'd be easier to read as `value.fallback || undefined` since it omits a ternary.

lib/renderJs.js
```diff
@@ -0,0 +1,33 @@
...
+        : value,
+    2
+  );
+}
+
+function renderJs(data) {
+  let output = '';
+  if (data.gesso) {
+    const gessoCategories = Object.keys(data.gesso);
```
gustavderdrache: Why not use [`Object.entries()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) here?

kmonahan: Because I remembered `Object.keys` but not `Object.entries()`. 😛 Will fix.

lib/renderJs.js
kmonahan: Yeah, good call. Will fix.

kmonahan:
Updated!

gustavderdrache (approved):

js/src/modules/_MobileMenu.es6.js
```diff
@@ -23,7 +24,7 @@ class _MobileMenu {
     menuButtonClass = 'mobile-menu-button--menu', // Class name for generated menu button
     closeButtonClass = 'mobile-menu-button--close', // Class name for generated close button
     arrowButtonClass = 'mobile-menu__subnav-arrow', // Class name for the subnav toggle
-    mobileMenuBreakpoint = '(max-width: 699px)', // Breakpoint to switch between mobile + original menu
+    mobileMenuBreakpoint = `(max-width: ${BREAKPOINTS['mobile-menu']})`, // Breakpoint to switch between mobile + original menu
```
coreylafferty, line 27: @kmonahan Should these be normal apostrophes around the variable?  

And is it feasible / a good idea to switch this from using max-width to min-width?  Originally we had a mobile-menu min-width breakpoint of 700px in the Sass, but set it to max-width 699 in the JS to match, but I don't think we use a CSS breakpoint for anything anymore.  Since we use min-width almost exclusively, it might be good to use it here unless something about the functionality needs max-width.

coreylafferty (changes requested):

js/src/modules/_MobileMenu.es6.js
kmonahan: 1) No, they should not. That's a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).
2) I used max-width deliberately here because the way the mobile menu works in Gesso, it's an addition to the default functionality. The mobile menu is created via JS and feels to me like a form of progressive enhancement. In particular, https://github.com/forumone/gesso/blob/8.x-3.x/js/src/modules/_MobileMenu.es6.js#L204 makes more sense to me using max-width, rather than writing a script to implement that functionality and then to undo it on larger screens. I agree that defaulting to min-width unless there's a reason to do otherwise is the right rule of thumb, but since the intention behind that, as I understand it, is to avoid serving unnecessary code to smaller screens with limited bandwidth, I will deviate from that in places where min-width results in more code all around.

coreylafferty (approved):

PR #332, opened 2020-02-06
adds important option to vertical spacing mixin and sets headers to u…

coreylafferty:
…se it by default


coreylafferty:
H1-H6 were not actually receiving their proper top margin when preceded by a <p> tags, since it comes later in the source order.  This add an !important option to the vertical spacing mixin so such cases can be overridden.

sbrack8t (approved):
Looks good to me.

kmonahan (approved):
👍

PR #333, opened 2020-02-07
Add back to top

kmonahan:
Adds a sticky back to top button (@dcmouyard, this is going to look really familiar). This provides the similar functionality to the Back to Top Drupal module but with the ability to customize the markup as needed (and without the need to install a module).

As part of this work, I also tweaked how the Pattern Lab CSS works. I replaced the existing sass glob module with one that supports nested globs, so that pattern-lab.scss can import everything in styles.scss, with the variable `$pattern-lab-overrides` set to true. This is so that a developer can add pattern lab specific overrides to a component where needed in the same sass file as the rest of the component styling, without the pattern-lab specific styling being included in the theme.


images/_sprite-source-files/arrow.svg
```diff
@@ -0,0 +1 @@
+<svg viewBox="0 0 19 12" xmlns="http://www.w3.org/2000/svg"><path d="m2.3 12 7.5-7.5 7.5 7.5 2.3-2.3-9.7-9.7-9.7 9.7 2.3 2.3z" fill="currentColor"/></svg>
```
dcmouyard: [nitpick] The `fill="currentColor"` attribute isn’t needed since we’re handling this in the icon component styles.

js/src/modules/_back-to-top.es6.js
```diff
@@ -0,0 +1,49 @@
+/**
```
dcmouyard, line 1: Why is the filename for this module different than others in this directory? (e.g., `_back-to-top.es6.js` instead of `_BackToTop.es6.js`?

source/_meta/_00-head.twig
```diff
@@ -23,5 +22,5 @@
     <!-- End Pattern Lab -->
 
   </head>
-  <body class="{{ body_class }}">
+  <body class="{{ body_class }}" id="top">
```
dcmouyard, line 25: We also need to add this id to the `_02-basic-head.twig` template.

source/_patterns/04-components/back-to-top/back-to-top.twig
```diff
@@ -0,0 +1,5 @@
+<a href="#{{ top_element }}" class="back-to-top" title="{{ text }}" aria-hidden="true">
+  <svg class="back-to-top__icon" role="img" aria-label="{{ text }}">
+    <use xlink:href="{{ image_path }}/sprite.artifact.svg#arrow"></use>
+  </svg>
```
dcmouyard: What’s the rationale behind baking this into the back-to-top component rather than using the icon component?

example:
```
{% include '@components/icon/icon.twig' with {
  'modifier_classes': 'icon--back-to-top',
  'is_hidden': true,
  'icon_name': 'arrow'
} %}
```
We would also need to add `aria-label` to the anchor tag if we went with this approach.

dcmouyard (changes requested):
Awesome work! I especially love that you found a fix for the glob importing and came up with the `$pattern-lab-overrides` trick!

js/src/modules/_back-to-top.es6.js
kmonahan: Because it's not a class, and the others are.

images/_sprite-source-files/arrow.svg
kmonahan: Fair enough. Will update.

source/_meta/_00-head.twig
kmonahan: Ah, didn't know about that one. Will update.

source/_patterns/04-components/back-to-top/back-to-top.twig
kmonahan: I did consider using the icon component here (and in fact started to write it that way) but ultimately decided not to because I didn't want the associated styling from the .icon class and I wanted slightly different markup. If the icon component had been larger/more complex, I would have gone with the variant class approach, but here, setting new styling to override the previous styling felt like overkill.

kmonahan:
Updated!

kmonahan:
@coreylafferty , where did you land on this one? Are you requesting changes (namely moving the images_path variable into a global preprocess) prior to merge?

coreylafferty (changes requested):
@kmonahan 
1.  Let's remove the loading of styles.css from 00-head.twig and 02-basic-head.twig, since we're now loading all the styles into pattern-lab.css.
2. If it's low effort, I think it'd be good to make image_path globally available.
3.  Should we rename the image variable to be gesso_image_path?  I ran a quick google search of 'drupal image_path' and a few things did come back, so making the name more specific might avoid conflicts.

kmonahan:
_Finally_ getting back to this one! All three changes made.

dcmouyard (approved):

source/_meta/_00-head.mustache
```diff
@@ -5,7 +5,6 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width" />
 
-    <link rel="stylesheet" href="../../../../css/style.css?{{ cacheBuster }}" media="all" />
```
coreylafferty: @kmonahan This isn't our styles.css file -- it's a file from pattern lab core.  I think we should leave this in.

coreylafferty (changes requested):
See comment about _00-head.mustache

PR #334, opened 2020-02-08
Update component generator

### LINKED ISSUE #331: component creation script generated MD file fails linting
Recommend replacing 'Do you want documentation (markdown)?" question with a title prompt that specifies any special requirements to pass linting (capitalization, etc.).

kmonahan:
Closes #331. Updates the component generator to always create markdown files and to prompt for a human-readable title, which is used in the markdown file. The default value for the title is the component name with dashes converted to spaces and each word capitalized, but the user can override this by entering their own title.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,50 +1,51 @@
 module.exports = {
-  "root": true,
-  "parserOptions": {
-    "ecmaVersion": 6,
-    "sourceType": "module"
+  root: true,
+  parserOptions: {
+    ecmaVersion: 6,
+    sourceType: 'module',
   },
-  "env": {
-    "es6": true,
-    "node": true,
-    "browser": true
+  env: {
+    es6: true,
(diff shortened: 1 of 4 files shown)


gustavderdrache:
How much effort do we want to put into updating this? I think this would be a great time to make the function async to avoid the promise callback chaining (and cut down on the `*Sync` methods in fs, which is always a plus).

kmonahan:
> How much effort do we want to put into updating this? I think this would be a great time to make the function async to avoid the promise callback chaining (and cut down on the `*Sync` methods in fs, which is always a plus).

I think a larger refactor is worth doing. However, I'm inclined to make it a separate issue + PR rather than address it here, so that the fix to #331 isn't blocked.

gustavderdrache (approved):
Fair point on the separate refactoring.

coreylafferty (approved):

PR #335, opened 2020-02-08
Add options to set button border width and color

kmonahan:
Adds options to set border width and color for a button style, for use in creating outline-style buttons, e.g. https://designsystem.digital.gov/components/button/.

### DIFF
--- a/source/_patterns/00-config/01-mixins/_mixins.button.scss
+++ b/source/_patterns/00-config/01-mixins/_mixins.button.scss
@@ -1,6 +1,7 @@
 // @file
 // Button styles
 
+$button-border-width: 0 !default;
 $button-border-radius: 0 !default;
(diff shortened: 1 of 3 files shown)


dcmouyard (approved):
Works great! 👍 

We might also want to add a disabled border color token:
```
disabled:
  background: grayscale.gray-2
  border: grayscale.gray-4
  text: grayscale.gray-4
```

kmonahan:
> Works great! 👍
> 
> We might also want to add a disabled border color token:
> 
> ```
> disabled:
>   background: grayscale.gray-2
>   border: grayscale.gray-4
>   text: grayscale.gray-4
> ```

Ah, yeah, that's a good idea! I'll do that in a follow-up PR (mostly because I don't have time right now). Will file an issue so I don't forget.

PR #337, opened 2020-02-11
Sidebar cols template and variable name

alecherryy:


sbrack8t:
This assumes that the sidebar is always first. Instead, should we include the option to flip the sidebar and content?

alecherryy:
Given the structure of the layout out of the box (first sidebar, main content, second sidebar), it made sense (to me) to have the template already set to `$sidebar-width 1fr `. As it stands now, the twig template presents the sidebar first, but the CSS sets it to be the larger fraction.

coreylafferty:
@alecherryy can you update the sidebar MD and YAML files as well with the proper variable names so everything is consistent?  

Adding a way to flip the sidebar easily might be nice, but I'm fine moving forward with this PR without it since this fixes bugs with the existing layout.

sbrack8t:
Gotcha: 👍

alecherryy:
Both `.yml` and `.md` file have the correct variable name.

coreylafferty:
@alecherryy sorry, I meant all the variables, not just the content one.  The sidebar_header_content variable is called header_content in the twig and md file, but sidebar_header_content in the yaml.  let's make it sidebar_header_content everywhere so that works and is consistent as well.

coreylafferty (approved):

PR #341, opened 2020-02-21
updates link state order to follow LVHFA

coreylafferty:
Link states should follow the Lord Vader Hates Furry Animals (Link, Visited, Hover, Focus, Active) order so that state transitions (like color) don't conflict under certain circumstances (e.g., you have a unique visited color defined).


ghost:
Not a comment on the commit, but I haven't seen that mnemonic before - that is fantastic.

kmonahan (approved):

PR #346, opened 2020-03-19
adjusted checkbox styles and template for cases where label is invisi…

coreylafferty:
…ble or before the input


kmonahan (approved):
lgtm!

PR #348, opened 2020-03-19
increased nesting depth, removed qualifying type check

coreylafferty:
This increases the nesting limit from 3 to 4 (I wouldn't be opposed to making it 5, but thought we'd start with baby steps).

This also removes the qualifying selector check.  There are already four places in the default Gesso styles where we had to explicitly ignore this rule, and I often run into it in new components.  I don't think it's unreasonable to want to style classes differently based on the elements (e.g., a.button vs button.button or div.align-right vs figure.align-right) and I also don't think this is something people often do without good reason.

### DIFF
--- a/.stylelintrc.yml
+++ b/.stylelintrc.yml
@@ -11,7 +11,7 @@ rules:
     - never
     - ignore:
       - 'inside-function'
-  max-nesting-depth: 3
+  max-nesting-depth: 4
   no-empty-first-line: true
   order/order:
     - - type: at-rule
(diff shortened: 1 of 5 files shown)


kmonahan (approved):
I agree with this. I've gone back and forth a bit, and argued for keeping it before, but I've ended up where you are. I don't think we're seeing people qualify selectors unnecessarily, and if there are places where that is happening, there's enough ambiguity around where it is and isn't necessary that it feels more like something that should be discussed in code review than something we have a clear stance on and should handle by linting. I'm also in favor of reducing false positives in the linter, because I don't want to create the impression that "add a comment to disable the rule" is the first tool you should reach for to fix linting errors. So I'm good with merging this!

dcmouyard (approved):
👍

PR #349, opened 2020-03-19
renamed content variables

coreylafferty:
Since Drupal's content variable is just named 'content', I've renamed all the pattern lab component variables that were named the same to something more specific to ensure that 'content' never accidentally bleeds through into a component and to avoid confusion.


dcmouyard:
@coreylafferty Do you have an example where the the Drupal content variable ever bled into the Pattern Lab variables? I typically prefer to have similar variables names across PL components if they usually have the same type of content. That way I can put in sample content for that variable in `data.yml`.

kmonahan:
Not opposed to this change, but not sure it's necessary either. I've deliberately given pattern lab variables the same name as Drupal variables before (if the name is sensible and not overly Drupal-specific) so that I can extend a PL template. While accidental bleedthrough is a different story, I'm wondering if it happens often enough to warrant using unique names for the variable in each component, or whether it's something that should be fixed on the project level.

coreylafferty:
@dcmouyard @kmonahan The one big time I ran into this, I was helping someone who had created a component with a {% if content %} {{ content }} {% endif %} block of code in it.  From their Drupal template, they'd called this component but not passed in a value for 'content' because they didn't want anything in that space in this instance.  They'd designed it so {{ content }} was just meant for some basic WYSIWYG text and if you didn't pass it in, that area wouldn't be populated.  However, since they didn't explicitly set 'content: false' in the Drupal template, what ended up being rendered was ALL the fields for that node, which isn't what they'd expected b/c they hadn't defined 'content' as anything... but Drupal's content variable bled through since it hadn't been overridden or set to 'false'.  

Personally, I prefer to have each component haven its own variable names to avoid any bleeding through or confusion.  In this particularly case, naming a lot of variables 'content' when Drupal very often has a monolithic variable with the same name seems like it could be an issue...  altho admittedly I've only seen a major issue with this the once.

kmonahan (approved):
This has been sitting a while and it sounds like 1) it'll definitely help in certain use cases (the one @coreylafferty outlined) and 2) those who haven't encountered that particular use case nonetheless don't have strong objections. Let's merge!

dcmouyard (approved):
👍

PR #350, opened 2020-03-19
adds captioned image component

coreylafferty:
This adds a 'captioned-image' component with left, right and centered variants that allows the caption to auto-size with the included image.  I've limited the `display: table` styles to non-screen readers so that the caption doesn't get read twice.

### DIFF
--- a/source/_patterns/04-components/captioned-image/_captioned-image.scss
+++ b/source/_patterns/04-components/captioned-image/_captioned-image.scss
@@ -1,0 +1,27 @@
+// @file
+// Component: Captioned Image
+
+.captioned-image {
+  @media not speech {
+    display: table;
+    figcaption {
+      caption-side: bottom;
+      display: table-caption;
+    }
+  }
+}
+
+.captioned-image--center-aligned {
+  margin-left: auto;
+  margin-right: auto;
+}
+
+.captioned-image--left-aligned {
+  float: left;
+  margin-right: rem(gesso-spacing(3));
+}
+
+.captioned-image--right-aligned {
+  float: right;
+  margin-left: rem(gesso-spacing(3));
+}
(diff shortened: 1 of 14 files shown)


dcmouyard (commented):
@coreylafferty I’m still hearing the photo alt text and caption twice in VoiceOver on Mac OS X in Firefox and Chrome. In Safari the caption is read twice, but not the alt text.

I’m leaning towards blocking this merge until that’s fixed. That said, it’s not _breaking_ the experience for screen reader users, it’s just making it a bad user experience.

coreylafferty:
@dcmouyard do you have any thoughts on why you'd be hearing it twice when I've set the CSS we've discussed to 'NOT speech' in the media query?  Should I set it to 'screen and print' instead?  Are we even sure it's this CSS that's the problem and not just figcaptions in general?

source/_patterns/04-components/captioned-image/_captioned-image.scss
```diff
@@ -0,0 +1,27 @@
+// @file
+// Component: Captioned Image
+
+.captioned-image {
+  @media not speech {
+    display: table;
+    figcaption {
+      caption-side: bottom;
+      display: table-caption;
+    }
```
dcmouyard: I think there should be an empty line between `display:table;` and `figcaption {`, but I’m not sure if that’s documented in our coding standards.

Any reason not to add a `captioned-image__caption` class on the `figcaption` element and use the class selector here instead?

dcmouyard (commented):
After digging into the screen reader side of thingsa bit more, it looks like there’s a VoiceOver bug that reads out the figure and caption twice, even without `display: table` being set.
https://www.powermapper.com/tests/screen-readers/labelling/img-figcaption/

So perhaps we can merge this in after all.

dcmouyard (approved):
👍

PR #355, opened 2020-03-24
JS utils animations - slide and fade

alecherryy:


kmonahan (approved):
Code looks good to me. (I've seen it in action on SonoSite too.) One suggestion: since you are so close to JSDoc already, bring your comment blocks in line with the standard (following https://jsdoc.app/about-getting-started.html). I don't think F1 has a hard rule of using JSDoc, but I find it helpful to follow, so my comments follow a predictable format. I think you just need to:

- Remove the 'Function: fadeIn.' part of each comment and start with the function description
- Include the type for each parameter
- Follow the formatting JSDoc uses (add a dash between the parameter name and description, capitalize the first letter in the description, and add a period at the end.)

Approving anyway though, because reformatting comments is _definitely_ not merge-blocking.

coreylafferty (changes requested):
@kmonahan do you think documentation should be added to the JS README showing how to use these?  

@alecherryy can you address KJ's documentation changes and possibly add some documentation to the js/README file?

kmonahan:
@coreylafferty , I did not call that out in my review intentionally, because I think each of these functions is single-purpose and named clearly, so I don't see a need for more in-depth documentation. I'm not opposed to adding them (certainly not ever going to reject a PR because of too much documentation), but I'm happy to merge this one without that.

alecherryy (commented):
I have updated both files to include JSDoc comments.

alecherryy:
@coreylafferty @kmonahan One thing I thought of, to make sure the slide functions work seamlessly, the target element needs to have both `margin` and `padding` equal to 0. Could that be something to add to the either the JSDoc or the README? I am thinking the former, rather than the latter.

kmonahan:
> @coreylafferty @kmonahan One thing I thought of, to make sure the slide functions work seamlessly, the target element needs to have both `margin` and `padding` equal to 0. Could that be something to add to the either the JSDoc or the README? I am thinking the former, rather than the latter.

Good flag. Yes, I think that would be a useful thing to add to the JSDoc.

PR #357, opened 2020-04-17
Enforce LVHFA order

### LINKED ISSUE #330: Enforce LVHFA order
The stylelint-selector-pseudo-class-lvhfa is in place in .stylelintrc.yml, which suggests that Stylelint _should_ be enforcing our preferred ordering for link properties. In practice, however, I'm not getting errors when I reorder properties on a project using Gesso.

kmonahan:
Fixes #330 by creating a stylelint plugin to lint for LVHFA order in link styles. The new plugin differs from the stylelint-selector-pseudo-class-lvhfa plugin that was in use previously in that it lints for ordering when various link states have different styles. So, for example, if you undo the change made in https://github.com/forumone/gesso/pull/341, you should now see linting errors, whereas previously something like
```
  &:hover,
  &:focus {
    color: red;
 }

 &:visited {
   color: blue
 }
```
would pass linting.


dcmouyard (approved):
Works great! 👍

PR #360, opened 2020-04-24
Refactor component creation script

### LINKED ISSUE #336: Refactor component script to use async
Avoids promise callback chaining and reduces `*Sync` methods in fs.

kmonahan:
Closes #336 . Reworks parts of the component generator script to use async/await instead of fs.*Sync methods and promise chaining, as was originally suggested in #334 . No changes to the functionality from an end-user perspective.

### DIFF
--- a/lib/component.js
+++ b/lib/component.js
@@ -6,97 +6,89 @@ const inquirer = require('inquirer');
 const path = require('path');
 const mkdirp = require('mkdirp');
 
-const isDirectory = source => fs.lstatSync(source).isDirectory();
+const isDirectory = async source => {
+  const stats = await fs.promises.lstat(source);
+  return stats.isDirectory();
+};
 
-const getDirectories = source => {
-  return fs
-    .readdirSync(source)
+const getDirectories = async source => {
+  const directoryFiles = await fs.promises.readdir(source);
+  return directoryFiles
     .map(name => path.join(source, name))
     .filter(isDirectory);
 };
 
-init();
-
-function init() {
-  const patternSrc = `${process.cwd()}/source/_patterns/`;
-  const patternDir = getDirectories(patternSrc);
-
+async function getMachineName() {
   const questions = [
     {
       type: 'input',
-      name: 'component_name',
+      name: 'componentName',
       message: 'What is the name your component?',
       filter: machineName,
     },
   ];
-  inquirer
-    .prompt(questions)
-    .then(answers => {
-      const componentName = answers.component_name.trim();
-      const defaultComponentTitle = humanName(componentName);
-      const questions = [
-        {
-          type: 'input',
-          name: 'component_title',
-          message: 'What is the human-readable title of your component?',
-          default: defaultComponentTitle,
-        },
-        {
-          type: 'list',
-          name: 'component_folder',
-          message: 'Component Location',
-          choices: patternDir.map(item => path.basename(item)),
-        },
-        {
-          type: 'input',
-          name: 'component_folder_sub',
-          message: 'Include subfolder or leave blank',
-        },
-      ];
-      return inquirer
-        .prompt(questions)
-        .then(answers =>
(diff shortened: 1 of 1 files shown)


lib/component.js
```diff
@@ -6,97 +6,89 @@ const inquirer = require('inquirer');
...
+      message: 'Component Location',
+      choices: patternDir.map(item => path.basename(item)),
+    },
+    {
+      type: 'input',
+      name: 'componentFolderSub',
+      message: 'Include subfolder or leave blank',
+    },
+  ];
+  return await inquirer.prompt(detailedQuestions);
```
gustavderdrache: If you're just returning a promise, you don't need to `await` it here (under the hood, it forces two yields: the first is the `await` inside this function, followed by any callers' `await`).

lib/component.js
```diff
@@ -6,97 +6,89 @@ const inquirer = require('inquirer');
...
+}
+
+async function init() {
+  const patternSrc = `${process.cwd()}/source/_patterns/`;
```
gustavderdrache: Let's clean this up to use `path.join` instead of concatenation.

lib/component.js
```diff
@@ -6,97 +6,89 @@ const inquirer = require('inquirer');
...
-    });
+  console.log(output);
+  const confirm = [
+    {
+      type: 'confirm',
+      name: 'confirm',
+      message: 'Is this what you want?',
+    },
+  ];
+  const confirmation = await inquirer.prompt(confirm);
```
gustavderdrache: I'd probably rename variables so that you're able to use destructuring:

```js
const prompt = [...];
const { confirm } = await inquirer.prompt(prompt);
```

lib/component.js
```diff
@@ -113,22 +105,25 @@ function humanName(name) {
...
     mkdirp(location)
-      .then(() => {
-        const filesArray = ['scss', 'twig', 'yml', 'md'];
-        filesArray.forEach(function(file) {
-          makeComponentFile(componentName, componentTitle, location, file);
-        });
-        console.log(`${componentTitle} created`);
+      .then(async () => {
+        const filesArray = ['scss', 'twig', 'yml', 'md'].map(
+          async ext =>
```
gustavderdrache: The promise returned by `makeComponentFile` isn't being used by this arrow function, so it's probably simple to remove the `async`/`await` here (same reason as mentioned in an earlier comment - the extra `await` forces an unnecessary yield).

lib/component.js
```diff
@@ -113,22 +105,25 @@ function humanName(name) {
...
     console.log('Component directory already exists');
   } else {
     mkdirp(location)
-      .then(() => {
-        const filesArray = ['scss', 'twig', 'yml', 'md'];
-        filesArray.forEach(function(file) {
-          makeComponentFile(componentName, componentTitle, location, file);
-        });
-        console.log(`${componentTitle} created`);
+      .then(async () => {
```
gustavderdrache: Why not `await mkdirp(location)`? It'll save you some nesting.

lib/component.js
```diff
@@ -163,3 +158,5 @@ title: ${componentTitle}
     }
   });
 }
+
+init();
```
gustavderdrache: Remember to `.catch` any async errors thrown by this function. Node will track rejected promises, but it's fairly inconsistent - right now it's a warning, but eventually it'll be a hard error.

gustavderdrache (approved):
It looks like there are a few superfluous `await`s that can be removed, and I've left comments on a few minor style nits, but there's really nothing major as far as I'm concerned.

Overall, a very nice cleanup!

kmonahan:
Thanks, @gustavderdrache ! @coreylafferty, fyi, I'd like to make the changes suggested here before this one gets merged.

kmonahan:
Changes made! This is ready to merge (unless anyone else has feedback).

PR #363, opened 2020-05-15
Add option to configure disabled button border

### LINKED ISSUE #340: Add disabled border token
As suggested in #335:

```
disabled:
  background: grayscale.gray-2
  border: grayscale.gray-4
  text: grayscale.gray-4
```

kmonahan:
Closes #340 . This was originally suggested in #335, so that button borders can be configured when disabled the same way they can in their base state and on hover.


dcmouyard (approved):
Works like a charm! 👍

PR #364, opened 2020-05-15
Switch Babel config method to babel.config.json

### LINKED ISSUE #361: Consider switching to babel.config.json
Currently, Gesso includes a .babelrc file for Babel configuration. As of Babel 7, project-level configuration should be in babel.config.json instead, as .babelrc won't be applied to node_modules. (See https://babeljs.io/docs/en/configuration and https://github.com/babel/babel-loader/issues/171#issuecomment-494614098 for reference.) Most of the time, this doesn't make any difference, since node_modules are excluded from the babel_loader in the webpack config anyway. However, some modules (e.g. https://swiperjs.com/get-started/) can be imported as ES6 modules and require transpiling for IE11. Switching to babel.config.json would allow a dev to whitelist modules as needed in the webpack config.

kmonahan:
Closes #361. As discussed in the issue ticket, babel.config.json is Babel's recommended config file format if you want to configure settings on a project level, including transpiling node modules. This change won't impact node modules initially, since they're excluded from the babel loader in the webpack config, but will make it easier for devs to transpile third-party modules when they need to. (Example: SwiperJS.)

You'll need to run `f1 build` to test, but you shouldn't notice anything behaving differently.


dcmouyard:
@kmonahan I ran into a problem due to the `services/gesso/Dockerfile` in my test environment still referring to `.babelrc`. I thought we had switched to having Gesso control its own Dockerfile.

Is this a potential problem, or is this due to me using an old version of webstarter as my local test environment?

kmonahan:
At first glance, that sounds like a problem with your test environment since, yes, Gesso uses its own Dockerfile these days. Do you get the same issue on a fresh project?

dcmouyard (approved):
After getting a new dev environment up & running, this PR works like a charm!

👍

PR #369, opened 2020-06-24
Fixes to slide toggle

### LINKED ISSUE #368: slide.es6.js seems to sometimes misfire
I've applied the slidetoggle animation for slide.es6.js to something and on click it occasionally skips the slide, simply showing or hiding the element immediately.  

Discussed this with @kmonahan - I can provide a repo and branch with an example for testing when this gets looked at.

kmonahan:
Closes #368 . This fixes an issue where, when slideToggle() was used, the target element sometimes just appeared or disappeared without the transition. It also modifies the script so that if you try to call slideToggle again before the previous toggle is finished, the call will be ignored. This is so that if you try to rapidly click on something that calls slideToggle, it won't start closing the slider before it's finished opening.

I went with the custom event route here instead of disabling the trigger button so that all of the functionality is contained within the script here. That makes the code here a bit more complicated but means that I could use it on Corey's test site without modifying anything outside of _slide.es6.js, so I think it'll be easier for other devs to drop into their projects this way.

@alecherryy , especially want your review on this one since you're most familiar with these scripts.

### DIFF
--- a/js/src/modules/_slide.es6.js
+++ b/js/src/modules/_slide.es6.js
@@ -5,11 +5,13 @@
  * elements.
  */
 
+import 'custom-event-polyfill';
+
 /**
  * Slides target element up and out view.
  *
  * @name slideUp
- * @param {string} target - The element sliding up.
+ * @param {HTMLElement} target - The element sliding up.
  * @param {integer} duration - The duration of the animation, with default value 500.
  */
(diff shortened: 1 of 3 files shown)


alecherryy (approved):
Looks great. Ready to merge.

PR #382, opened 2020-07-29
Updates to prepare for Drupal 9

kmonahan:
Includes work from the following PRs:

- https://github.com/forumone/gesso/pull/375
- https://github.com/forumone/gesso/pull/376
- https://github.com/forumone/gesso/pull/378
- https://github.com/forumone/gesso/pull/380
- https://github.com/forumone/gesso/pull/381

You can also see the various Drupal 9 compatibility tickets at https://github.com/forumone/gesso/issues?q=is%3Aissue+is%3Aclosed+label%3Adrupal9

Work is backwards-compatible with Drupal 8.9, which is why I'm proposing merging it into the existing branch rather than maintaining separate Drupal 8 and Drupal 9 major branches. However, it is not compatible with versions of Drupal 8 *older* than 8.9. I can't think of a use case where you would be starting with a fresh install of Gesso but not the latest Drupal 8 release, but if that is a concern, then we could keep them separate. In that case, I would recommend that 9.x-3.x (or 8.x-4.x if we'd rather) become the new base branch.

The breaking changes (for older version of Drupal 8) are the new format of declaring core compatibility and the addition of module dependencies in the theme info file. (So that *finally* you won't be able to enable Gesso without enabling the Components module and get a white error screen.) Changes overall are honestly pretty minimal -- a couple replacements to deprecated code and some formatting that got flagged by Drupal's automated patch.


dcmouyard (commented):
I can confirm that this PR works in Drupal 8.9.2. 👍 

I’ll take a quick stab at fixing the merge conflict.

coreylafferty (approved):

coreylafferty:
Closing as this was merged into 4.x

PR #387, opened 2020-08-20
Support Background Images Formatter module

### LINKED ISSUE #384: Consider alternate background image implementations
Currently, the default hero component renders the background image using an inline style tag. Better would be to have that CSS be generated as a style tag in the head. (See discussion in #f1-gesso for full background.)

Work needed is to test and standardize on a module that can do this (or write our own field formatter, although there seem to be plenty of options to try first), add the module to the Gesso README, and update the Pattern Lab component to assume use of the module. We might also want to add it as a dependency of the theme in the 9.x-3.x branch if it's required rather than optional and potentially also open a PR on the web starter generator to include the module as one of the ones that gets installed when Gesso does.

Possible modules:

- [Background Image](https://www.drupal.org/project/background_image)
- [Simple Background image formatter](https://www.drupal.org/project/background_image_formatter)
- [Background Images Formatter](https://www.drupal.org/project/bg_image_formatter)

kmonahan:
Resolves #384 . As discussed in that ticket and in Slack, sites may need or want to use a module that formats background images via CSS inserted into the head rather than the inline style attribute. Changes here add Background Images Formatter as an optional module to the README and modify the hero component so that the inline style attribute is added only if an image URL is passed into the component. This allows a dev to omit the inline attributes if they're using the module.

These changes don't disallow the style attribute approach, just set Gesso up to work either way. (Or mixed, if you want to use a URL in the pattern lab demo but the module on the Drupal side.) I think being able to use responsive images for hero backgrounds or other potentially very large images is beneficial even if a site doesn't need to pass a security scan that disallows style attributes.


dcmouyard (approved):
Changes look good to me. 👍 

I rebased this PR to the latest `8.x-3.x` to fix the merge conflict.

PR #388, opened 2020-08-24
Add BrowserSync for live reloading

### LINKED ISSUE #362: Use utility like livereload to refresh styles and js automatically when gesso builds

kmonahan:
Resolves #362 . Adds [Browsersync](https://browsersync.io/) to Gesso for automatic CSS injection when CSS is updated and page reloading when JS or Twig files are updated. This is a draft PR because I'll also need to open a PR on the f1 cli tools in order to get this working correctly, so it's not ready to use quite yet.

## How It'll Work
Running the gesso gulp task starts Browsersync. Browsersync proxies a URL, so instead of viewing the site at localhost:8080, you'll view it at localhost:3000. (The terminal output includes a link to `http://localhost:3000/themes/gesso/pattern-lab/?p=all`.) When front-end build tasks finish, Browsersync will update any browsers using the proxy URL. One nice thing about Browsersync is that it can synchronize multiple browsers at the same time, so if you're doing cross-browser testing and have Firefox, Safari, and Chrome all open to the same page (using the localhost:3000 URL), they'll all get updated in sync. Browsersync also provides an external IP for accessing the dev page from another device, i.e. a phone.

I have the start path defaulting to Pattern Lab because that's where I imagine it'll be the most used, since that's where we write most of our CSS and Twig. However, I've used Browsersync before with regular CMS pages too, and http://localhost:3000 will display whatever http://localhost:8080 would. One caveat is that I think you're going to need CSS to be unminified by Drupal for Browsersync to be able to inject the updated file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,3 +4,6 @@ COPY package*.json ./
 RUN if test -e package-lock.json; then npm ci; else npm i; fi
 
 COPY . .
+
+EXPOSE 3000
+EXPOSE 3001
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const { dest, lastRun, parallel, series, src, watch, task } = require('gulp');
+const browserSync = require('browser-sync').create();
 const patternLabConfig = require('./pattern-lab-config.json');
 const patternLab = require('@pattern-lab/core')(patternLabConfig);
 const postcss = require('gulp-postcss');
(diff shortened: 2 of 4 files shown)


ghost:
Very cool idea. Would there be a way to turn this off if needed? I love BrowserSync for some testing, but I also use Polypane and BrowserSync doesn't work well with Polypane as it has its own auto-refresh and synced scrolling so would be nice to turn on/off as needed. Thanks!

PR #390, opened 2020-09-02
switched component variants to use pseudo patterns

coreylafferty:


dcmouyard (approved):
I was hesitant about this change since I thought I would need to revert all of this work on my projects since I use [pattern states](https://patternlab.io/docs/using-pattern-states/) to track which components and variants are done.

It turns out, however, that I can just add `component~variant.md` files to adjust the names and pattern states. 👍

dcmouyard:
Would it be worth it to add markdown files for each variant? I’m going to have to do that on my next project after this PR has been merged.

PR #392, opened 2020-09-28
Add modifier classes to nav component

dcmouyard:
This PR adds the option to add modifier classes to the nav component.


kmonahan (approved):
I think we missed this one in our review just now, but 👍 from me.

Issue #408, opened 2020-10-21
Add default styles for align classes (align-left, align-right, etc.) and create filter-caption.html.twig


coreylafferty:
filter-caption.html.twig should be tied to captioned-image component in source.

PR #409, opened 2020-10-23
Refactor utility classes

dcmouyard:
This PR:
* Adds a new `escape-number()` Sass function that will replace decimals in numbers with hyphens. This is useful if you want decimal numbers to work in class names.
* Breaks out classes in `_utility-classes.scss` into separate partials inside a new `utility-classes` folder
* Adds alignment utility classes (which would be useful to tackle #408)
* Adds full-width utility class, which will break elements outside of `l-constrain`
* Adds spaced utility classes, which allows one to easily add vertical spacing to a component without having to create new variants or states for each component


coreylafferty (changes requested):
@dcmouyard I made a few additions to this in a separate PR to your branch.  Let me know what you think.

https://github.com/forumone/gesso/pull/411

PR #411, opened 2020-11-09
added caption image template and tweaked utility styles

coreylafferty:
- Added the classes that Drupal uses out of the box for alignment (e.g., if you add an image to the WYSIWYG): align-left, align-right, align-center
- Added a template for captioned images from the the WYSIWYG that makes use of our captioned-image component.  Added the role="group" property to this component if a caption is present.

### DIFF
--- a/source/_patterns/04-components/captioned-image/captioned-image.twig
+++ b/source/_patterns/04-components/captioned-image/captioned-image.twig
@@ -3,7 +3,7 @@
   modifier_classes ? modifier_classes,
 ]|join(' ')|trim %}
 
-<figure class="{{ classes }}">
+<figure class="{{ classes }}"{% if caption %} role="group"{% endif %}>
   {{ media }}
   {% if caption %}
     <figcaption class="captioned-image__caption">{{ caption }}</figcaption>
--- a/source/_patterns/utility-classes/_alignment.scss
+++ b/source/_patterns/utility-classes/_alignment.scss
@@ -2,18 +2,29 @@
 // Alignment utility classes
 
 // Align items
-.u-align-left {
+.u-align-left,
+.align-left {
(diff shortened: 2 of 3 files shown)


source/_patterns/04-components/captioned-image/captioned-image.twig
```diff
@@ -3,7 +3,7 @@
   modifier_classes ? modifier_classes,
 ]|join(' ')|trim %}
 
-<figure class="{{ classes }}">
+<figure class="{{ classes }}"{% if caption %} role="group"{% endif %}>
```
dcmouyard, line 6: Why is `role="group"` added here?

source/_patterns/utility-classes/_alignment.scss
```diff
@@ -2,20 +2,31 @@
...
 }
 
+.u-align-center,
+.align-center {
+  margin-bottom: rem(gesso-get-map(gutter-width));
+  margin-left: auto;
+  margin-right: auto;
+  margin-top: rem(gesso-get-map(gutter-width));
+  text-align: center;
+}
```
dcmouyard: Why are top & bottom margins added here?

I don’t think `text-align: center` should be here. I often see centered images with caption, but the caption is still left-aligned.

dcmouyard (approved):
I’ve got a few questions, but nothing that would block merging this.

source/_patterns/04-components/captioned-image/captioned-image.twig
coreylafferty: I added role=group because that's in the original stable theme's file (filter-caption.html.twig)

source/_patterns/utility-classes/_alignment.scss
coreylafferty: You're right -- I've removed those.

PR #412, opened 2020-11-17
Bug - <nolink> parent menu item

### LINKED ISSUE #410: Bug - <nolink> parent menu item
If a parent menu item is set to `<nolink>`, the current code is breaking on https://github.com/forumone/gesso/blob/5c02bd2944c3cc6791fe61fb830c28f8c53f5b90/js/src/modules/_Menu.es6.js#L127-L129

sbrack8t:
Fixes #410
- Check for valid 'A' or 'BUTTON' TAG
- Add styling for possible button as submenu trigger

### DIFF
--- a/js/README.md
+++ b/js/README.md
@@ -18,6 +18,11 @@ The styles for this menu are in `source/_patterns/04-components/menus/menu--main
 By default, the script is applied to a menu with the class `menu--main`. To use
 a different class, change the query selector in `primary-menu.es6.js`.
 
+The primary menu script does NOT work with menu items whose link is set to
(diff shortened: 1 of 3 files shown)


sbrack8t:
I added support for buttons but I'm not sure we should add support for the `<nolink>` options. We would need to include additional JavaScript to make the nolink (span) tab friendly

johnbburg:
I just had a project where this broke due to site admins using the <nolink> route. 
+1 on this being needed.

coreylafferty (approved):

Issue #365, opened 2020-05-22
Drupal 8 Responsive tables


johnbburg:
Drupal 8 core provides a responsive table feature. However, any theme would need to implement responsive styling for that feature's classes. Gesso does not currently. 

See https://drupalize.me/blog/201409/how-add-responsive-tables-content-drupal-8.

Views are easily set up this way, it's a bit of a manual process for tables defined in the textareas.

But It looks like we just need to implement some biolerplae styles for the following classes:

responsive-enabled
priority-high
priority-low
priority-medium

johnbburg:
~~I should have looked deeper into the responsive table feature in core first. It that this is driven entirely by JS.~~

Nevermind, this is some simple CSS. I think what I saw was something entirely different.

johnbburg:
Should be something like this from Bartik:

```
/**
* Responsive tables.
*/
@media screen and (max-width: 37.5em) { /* 600px */
th.priority-low,
td.priority-low,
th.priority-medium,
td.priority-medium {
display: none;
}
}
@media screen and (max-width: 60em) { /* 920px */
th.priority-low,
td.priority-low {
display: none;
}
}
```

PR #417, opened 2021-01-15
issue365 adds priority styles for view table cells

coreylafferty:


coreylafferty:
Closes #365

kmonahan (approved):
Ah, nice, never knew Drupal offered this.

PR #421, opened 2021-02-04
Fix original classes on menu item links

dcmouyard:


coreylafferty (changes requested):
I tested this by adding the menu_link_attributes module and ran into a WSOD error with 'The website encountered an unexpected error. Please try again later.' Here's what I tested and how to reproduce.

On _4.x_ branch:
- Install menu_links_attributes module.
- Add properties to a menu item (e.g., class and/or a target).
- Clear cache.
Result: class and target work as expected.

Switch to _menu-item-link-classes_ branch:
- ensure same properties are on the link
- clear cache
Result: Site crashes with 'The website encountered an unexpected error. Please try again later.' message.

Unless you've completely disabled caching locally, I think clearing the cache will be necessary for testing to pick up the file changes.

dcmouyard:
@azinck What was the initial problem this fix was for on the EPA project? I’m curious why we didn’t get the WSOD error.

I’ll check to see if the menu_links_attribute module causes the same error on EPA.

azinck:
@dcmouyard Without this change we were seeing Array to string conversion errors whenever displaying menu items.

azinck:
Without diving into this, it sounds like menu_links_attributes module is converting the item['original_link'].options.attributes.class property to a string, necessitating you wrapping it in an array...while on EPA that was coming through as an array already.  It's conceivable that there's some module or configuration on the EPA project that is causing it to come through as an array when it shouldn't be, but I'm not so sure about that.  

I also am skeptical of the initial version of this code (pre-change) since that code is written in a way that clearly expects item['original_link'].options.attributes.class to be a string, and the code ( `set link_classes = link_classes|merge([item['original_link'].options.attributes.class]) ` ) will just take the entire string (which could contain multiple classnames) and inject them into a single-element array with all the classnames on that one array element which is not really the intended structure (you'd normally want one classname per array element).  But I may be missing something here.

dcmouyard:
When I enabled the menu_link_attributes module locally on the EPA project and added a custom class to a menu item, I got the following error:

```
Twig\Error\RuntimeError: The merge filter only works with arrays or "Traversable", got "string" as second argument. 
in twig_array_merge() (line 52 of themes/epa_theme/source/_patterns/05-components/menu/menu.twig). 
```

This confirms what @azinck said about the menu_link_attributes module passing in a string incorrectly instead of an array. There’s a d.o bug on this: https://www.drupal.org/project/menu_link_attributes/issues/3193364

I think this PR should wait until that bug is fixed in the menu_link_attributes module since we use it quite a bit on projects.

coreylafferty:
@dcmouyard @azinck I've updated this PR with some code that @mrclay came up with for another project.  Instead of switching from assuming the class is an array to assuming it's a string, this works for either case.  In my testing this worked with and without the menu_link_attributes module.  Closes #444

coreylafferty (approved):

PR #428, opened 2021-02-12
Field variants

dcmouyard:
This PR:
* Updates markup & styles for the field component
* Adds the following field variants:
  * field--list
  * field--no-separator
  * field--simple
  * field--tight
* Updates Drupal field template to use new component, defaulting to `div` instead of `h3` for field labels


coreylafferty:
@dcmouyard When I create variants like this, I typically name them in a way to avoid confusion with class names created from actual field machine names.  For example, 'field--style-list' or 'field--display-list' instead of just 'field--list'.  That way it's less likely to run into naming collisions and makes it easier to see which class names are for specific fields and which are for different ways of displaying fields.  Not sure it's a necessity, but thoughts on a approach like that?

coreylafferty (approved):

PR #430, opened 2021-02-23
Fix tabindex of mobile menu links

dcmouyard:
The `_setTabIndex()` function wasn’t correctly setting the `tabindex` attribute on mobile menu links if the `elem` parameter had a NodeList inside an Array.

### DIFF
--- a/js/src/modules/_MobileMenu.es6.js
+++ b/js/src/modules/_MobileMenu.es6.js
@@ -164,7 +164,14 @@ class _MobileMenu {
   _setTabIndex(elem, tabIndex) {
     if (Array.isArray(elem)) {
       elem.forEach(function (item) {
-        item.tabIndex = tabIndex;
+        // Check if item is a NodeList.
+        if (Object.prototype.isPrototypeOf.call(NodeList.prototype, item)) {
+          for (let i = 0; i < item.length; i++) {
+            item[i].tabIndex = tabIndex;
+          }
+        } else {
+          item.tabIndex = tabIndex;
+        }
       });
     } else {
       elem.tabIndex = tabIndex;


js/src/modules/_MobileMenu.es6.js
```diff
@@ -164,7 +164,14 @@ class _MobileMenu {
   _setTabIndex(elem, tabIndex) {
     if (Array.isArray(elem)) {
```
sbrack8t, line 165: it might be cleaner to either check if the parameter is iterable (array or NodeList) then convert it to an array instead of doing the for loop
```
function isIterable(obj) {
  // checks for null and undefined
  if (obj == null) {
    return false;
  }
  return typeof obj[Symbol.iterator] === 'function';
}
```

coreylafferty (changes requested):
@dcmouyard can you take a look at Shawn's suggestion and incorporate if needed?  If you want to leave as-is, let me know and I'll retest.

js/src/modules/_MobileMenu.es6.js
dcmouyard: > it might be cleaner to either check if the parameter is iterable (array or NodeList) then convert it to an array instead of doing the for loop
> 
> ```
> function isIterable(obj) {
>   // checks for null and undefined
>   if (obj == null) {
>     return false;
>   }
>   return typeof obj[Symbol.iterator] === 'function';
> }
> ```

@sbrack8t This JS is over my head, and I’m not sure how I would implement this. Is the potential performance improvement worth it?

sbrack8t: For the number of potential items in the menu list there's probably not much of a performance gain. I'm fine leaving it as is.

coreylafferty (approved):

PR #436, opened 2021-03-19
Codacy integration

slucero:
This adds configuration for Codacy to scan the repo and new commits pushed to it.

### Local Execution

Codacy may be executed locally using the provided Docker Compose configuration as documented in the `docker-compose.yml` file.

#### PHPCS

```bash
docker-compose run phpcs
```

#### ESLint

```bash
docker-compose run eslint
```

#### Codacy

Review of all available Codacy CLI usage may be accomplished with:

```bash
docker-compose run codacy --help
```

All scanners may be completed using the following (An output file is recommended due to the amount of output.):

```bash
docker-compose run codacy analyze -o quality.txt
```

[A specific tool](https://docs.codacy.com/repositories-configure/codacy-configuration-file/#which-tools-can-be-configured-and-which-name-should-i-use), aside from the pre-configured ones noted above, may be executed using:

```bash
docker-compose run codacy analyze -t <tool>
```


timnolte:
@slucero for consistency with other projects would it make more sense to put the Codacy Docker setup in the `docker-compose.cli.yml`. That is what I did in the Gesso WP work. At the very least we should probably bring alignment between Gesso & Gesso WP in regards to this.

I can update my PR for Gesso WP to move it to the `docker-compose.yml`.

slucero:
@timnolte, my reasoning for putting it into `docker-compose.yml` instead of `docker-compose.cli.yml` was this project doesn't have any traditional running services that would typically live in the `docker-compose.yml` file. It doesn't matter much either way, but I put it here to keep it simpler in execution consistent with `docker-compose` directly without `f1 run` where the additional cli config file would have to be explicitly specified.

timnolte:
@slucero OK, I can get behind removing the requirement that the F1 CLI is required. I'm inclined to keep things, in Gesso WordPress, such that a Developer could use the F1 CLI for consistency between the theme development and site work. I'll make this change in my Gesso WP work. Thanks!

kmonahan:
Looks good! I did see a lot of noise from the files in `js/overrides`, because those are outside our webpack build setup and thus haven't been going through linting up until now. Given that we are also trying to make this the year we move away from jQuery, I wonder if we should ignore those files in codacy.yml for now until we can move them to vanilla JS, following our current coding standards. @sbrack8t , @coreylafferty , @dcmouyard , thoughts?

kmonahan (approved):

kmonahan:
Can we disable whatever Markdown linting is happening here? We've got an established convention for documenting Pattern Lab components and variables that's followed pretty consistently in Gesso (and some custom linting in Gesso already to lint those files). I don't think the things Codacy is flagging are necessary.

slucero:
> Can we disable whatever Markdown linting is happening here? We've got an established convention for documenting Pattern Lab components and variables that's followed pretty consistently in Gesso (and some custom linting in Gesso already to lint those files). I don't think the things Codacy is flagging are necessary.

@kmonahan, I've disabled this in 4e1f66c as part of the re-rolled configuration PR: #441.

slucero:
I re-rolled this PR into two separate PRs to introduce the Codacy configuration first, and the local execution configuration and documentation separately.

I'm closing this PR now in favor of #441 and #442.

PR #440, opened 2021-04-05
Address ESLint issues identified with Codacy

kmonahan:
Removes the placeholder annotations.js file but keeps the _annotations directory. I'm not sure whether we _need_ to keep the annotations directory, but it's consistent with how we're handling the _layouts directory.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -86,7 +86,7 @@ const compileStyles = () => {
       sass({
         includePaths: ['./node_modules/breakpoint-sass/stylesheets'],
         precision: 10,
-        importer: sassGlobImporter()
+        importer: sassGlobImporter(),
       })
     )
     .pipe(
--- a/source/_annotations/annotations.js
+++ b/source/_annotations/annotations.js
@@ -1,0 +1,9 @@
+var comments = {
+	"comments" : [
+		{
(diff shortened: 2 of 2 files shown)


coreylafferty:
@kmonahan What's the downside to keeping the annotations example file?  I kind of like it there as it gives you the pattern to follow if you want to add PL annotations.  Would it be better to just to have it as an empty JS file with a comment at the top with the pattern?  Or add a README file in there with a link to the PL documentation page around annotations?

kmonahan:
It's not so much as a functional downside (we'd have to disable linting on the annotations.js file, but that's easily done) so much as me not loving having an annotation that doesn't annotate anything. I really like the idea of a README with a link to the documentation, since that both helps people figure out what that directory is for and points them to info on the JSON and YML options.

The more I think about it, I kind of want to do that for the other PL directories too, even when the content in them is used by Gesso.

kmonahan:
@coreylafferty, given that, maybe we put the placeholder annotations file back for now with an `/* eslint-disable */` comment at the top, and then I'll open a new PR with READMEs.

kmonahan:
@coreylafferty Updated per discussion above

coreylafferty (approved):

PR #441, opened 2021-04-06
Add Codacy configuration files

slucero:
This PR is a re-roll of pull request #436 to only include configuration files to support online scanning with Codacy. A follow-up PR will add configuration for local execution.

PR #442, opened 2021-04-06
Codacy local execution

slucero:
This PR builds on top of #441 to support local execution of the [Codacy CLI](https://github.com/codacy/codacy-analysis-cli) for scanning outside of the PR process.

- [x] Add docker-compose configuration to run the Codacy CLI
- [x] Add common tool shortcuts for Codacy CLI execution
- [ ] Add README documentation about the role of Codacy and usage

### Local Execution

Codacy may be executed locally using the provided Docker Compose configuration as documented in the `docker-compose.yml` file.

#### PHPCS

```bash
docker-compose run phpcs
```

#### ESLint

```bash
docker-compose run eslint
```

#### Codacy

Review of all available Codacy CLI usage may be accomplished with:

```bash
docker-compose run codacy --help
```

All scanners may be completed using the following (An output file is recommended due to the amount of output.):

```bash
docker-compose run codacy analyze -o quality.txt
```

[A specific tool](https://docs.codacy.com/repositories-configure/codacy-configuration-file/#which-tools-can-be-configured-and-which-name-should-i-use), aside from the pre-configured ones noted above, may be executed using:

```bash
docker-compose run codacy analyze -t <tool>
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,9 +38,12 @@ services:
     <<: *codacy
     entrypoint: /opt/codacy/bin/codacy-analysis-cli analyze -t eslint
     command: ''
-    
+
   # Run StyleLint linting with `docker-compose run stylelint`.
   stylelint:
     <<: *codacy
     entrypoint: /opt/codacy/bin/codacy-analysis-cli analyze -t stylelint
     command: ''
+volumes:
+  ? gesso-pattern-lab
+  ? gesso-css
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,9 +1,18 @@
 version: "3.7"
 services:
   gesso:
-    build: .
+    build: ./
+    init: true
     volumes:
-      - "$PWD:/"
+      - ./images:/app/images:cached
+      - ./js:/app/js:cached
+      - ./source:/app/source:cached
+      - type: volume
+        source: gesso-pattern-lab
+        target: /app/pattern-lab
+      - type: volume
+        source: gesso-css
+        target: /app/css
 
   # Run all scans with `docker-compose run codacy`.
   # Target a specific tool with `docker-compose run codacy analyze -t <tool>`
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,7 +23,7 @@ services:
     working_dir: "${PWD}"
     command: analyze
     volumes:
-      - "$PWD:$PWD"
+      - ${PWD}:${PWD}:cached
       - /var/run/docker.sock:/var/run/docker.sock
       - /tmp:/tmp
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,3 +1,6 @@
+# This is the Compose file for command-line services.
+# Anything that doesn't need to be run as part of the main `docker-compose up'
+# command should reside in here and be invoked by a helper script.
 version: "3.7"
 services:
   gesso:
@@ -14,8 +17,9 @@ services:
         source: gesso-css
         target: /app/css
 
-  # Run all scans with `docker-compose run codacy`.
+  # Run all scans with `docker-compose run codacy` or `f1 run codacy`.
   # Target a specific tool with `docker-compose run codacy analyze -t <tool>`
+  # or `f1 run codacy analyze -t <tool>`.
   codacy: &codacy
     image: codacy/codacy-analysis-cli:latest
     environment:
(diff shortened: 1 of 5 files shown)


kmonahan:
@slucero , when I test this in a Drupal 8 sandbox, I'm seeing a couple issues. Running `f1 up` within the Gesso directory results in 
```
ERROR: for gesso  Cannot create container for service gesso: invalid volume specification: '/host_mnt/Users/kmonahan/Sites/d8-gesso-sandbox/services/drupal/public/themes/gesso:/:rw': invalid mount config for type "bind": invalid specification: destination can't be '/'
```
I get the same error when running `f1 gesso gulp build`.

The WP PR has a different setup for the gesso container in the theme's docker-compose.yml. Since I'm not seeing the issue in my WP sandbox, wondering if those changes need to be pulled in here: https://github.com/forumone/gesso-wp/pull/240/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R7-R23

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
+version: "3.7"
+services:
+  gesso:
+    build: .
+    volumes:
+      - "$PWD:/"
```
slucero: Make this match the volume mount configuration in https://github.com/forumone/gesso-wp/pull/240/files
```suggestion
      - ./images:/app/images:cached
      - ./js:/app/js:cached
      - ./source:/app/source:cached
      - type: volume
        source: gesso-pattern-lab
        target: /app/pattern-lab
      - type: volume
        source: gesso-css
        target: /app/css
```

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
+version: "3.7"
+services:
+  gesso:
+    build: .
```
slucero: Align this with https://github.com/forumone/gesso-wp/pull/240/files
```suggestion
    build: ./
    init: true
```

slucero:
I've made some changes for alignment with https://github.com/forumone/gesso-wp/pull/240/files, so this should be ready for re-testing.

kmonahan:
Initially, I saw the error
```
ERROR: Named volume "{'type': 'volume', 'source': 'gesso-pattern-lab', 'target': '/app/pattern-lab'}" is used in service "gesso" but no declaration was found in the volumes section.
    Error: Command failed: docker-compose up --build --detach
    Code: 1
```

I found one other difference between the Drupal and WP versions and pulled the volumes at the bottom of the WP file into here as well. That fixed the error.

kmonahan:
Testing locally (with the additional commit I made above), I'm seeing the following:
`f1 up`: works from the theme directory but does *not* actually start the site. I have to run `f1 up` outside the theme directory in order to get a local site up and running.
`f1 run gesso gulp build`: works from the theme directory but does *not* build the front-end assets so that Drupal can use them. I have to run the same command outside of the theme directory in order to see styles in Drupal.
`f1 drush cr`: does *not* work from the theme directory

So it seems the lack of errors on some commands is misleading, because they aren't actually _working_ in the sense of "doing what i think they are going to do."

There is a workaround (go up one directory level before running commands) but noting this because previously, I could run commands without needing to be in a specific directory within the project. -I will check whether the same behavior occurs on the WP side.- Confirmed that this is the same whether testing in Drupal or Wordpress.

kmonahan:
@coreylafferty , how much of an issue is the above? I do tend to run commands from the theme directory, but I imagine I could learn to not do that fairly easily. :)

timnolte:
@kmonahan so I think this is because we've added Docker into Gesso, which wasn't there before. This issue is partly to do with the F1-CLI/Docker looking in the current directory for Docker configuration, and then moving up the tree when it doesn't find it. So really these Docker configuration files are for Gesso WP starter development an not meant to be used within a project. The project level Docker configuration should be updated to support code linting on the entire project including Gesso,.

timnolte:
@slucero is it possible to have the Docker Compose configuration dropped by Web Start when Gesso is included in a project?

coreylafferty:
@slucero @timnolte seeing errors like the below when I try to run commands on a Windows PC.  I was NOT seeing these errors before upgrading to Docker 3.4.  However, when I downgraded and tested using Docker 2.5, while I did not see any errors, I don't think the commands were actually working.  I made changes to several PHP, JS and SCSS files that should have generated errors by codacy/phpcs, and yet the commands finished normally.  This is true of both the Drupal and WP PRs (i.e., https://github.com/forumone/gesso-wp/pull/254).  

```
Error response from daemon: OCI runtime create failed: invalid mount {Destination:::cached Type:bind 
Source:/var/lib/docker/volumes/12b75d1fafa83d8bbcbba1e7070baf0ddc0eab64cd054f97429f3938c80f3f80/_data Options:
[rbind]}: mount destination ::cached not absolute: unknown
```

timnolte:
> @slucero @timnolte seeing errors like the below when I try to run commands on a Windows PC. I was NOT seeing these errors before upgrading to Docker 3.4. However, when I downgraded and tested using Docker 2.5, while I did not see any errors, I don't think the commands were actually working. I made changes to several PHP, JS and SCSS files that should have generated errors by codacy/phpcs, and yet the commands finished normally. This is true of both the Drupal and WP PRs (i.e., [forumone/gesso-wp#254](https://github.com/forumone/gesso-wp/pull/254)).
> 
> ```
> Error response from daemon: OCI runtime create failed: invalid mount {Destination:::cached Type:bind 
> Source:/var/lib/docker/volumes/12b75d1fafa83d8bbcbba1e7070baf0ddc0eab64cd054f97429f3938c80f3f80/_data Options:
> [rbind]}: mount destination ::cached not absolute: unknown
> ```

@coreylafferty I suspect this may be due to how the `docker-compose.yml` file is setup with path references. @slucero I'm wondering if we need to re-evaluate how some paths are being specified in order to be fully cross-platform compatible?

PR #457, opened 2021-06-30
Add composer.json

kmonahan:
Adds a composer.json file to install PHP linting tools locally for IDEs. This is the Drupal-ified equivalent to https://github.com/forumone/gesso-wp/pull/242.


.editorconfig
```diff
@@ -0,0 +1,23 @@
...
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = space
+indent_size = 2
```
kmonahan, line 15: @coreylafferty These two lines are different from the WP version because Drupal coding standards use two spaces for indent: https://www.drupal.org/docs/develop/standards/coding-standards#indenting

PR #463, opened 2021-07-21
created new breakpoint mixins

coreylafferty:
@dcmouyard @kmonahan Here's my pass at adjusting the breakpoint mixins to handle multiple values and refactored how it handles max-width.  

I'm leaning towards NOT having it automagically subtract 1px when using max-width, partly b/c I couldn't figure out how to only apply it when using `gesso-breakpoint` values, but also because I'm not sure we want that to be the default behavior.  I'm still thinking that through, but if we do want it to be the default behavior, maybe we add a flag to disable/enable it?    

I didn't want that to hold up getting the basics of this in place, so let me know what you think of these changes and if you think we should continue down that path of the 1px subtraction, how you think that should best be implemented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,14 +308,17 @@ CSS and JavaScript, you should use simple strings or numbers.
 ### Width Based Media Queries
 
 Gesso uses custom mixins to specify viewport width based media queries:
-`breakpoint`, `breakpoint-max` and `breakpoint-min-max` for min-width,
-max-width or a combination of both, respectively. Each mixin takes one or two
+* `breakpoint`: min-width queries
+* `breakpoint-max`: max-width queries
+* `breakpoint-min-max`: queries with both a min and max width
+
+Each mixin takes one or two
 width parameters, which can be a straight value (e.g., 800px, 40em) or a design
 token value called using the `gesso-breakpoint` function (e.g.,
 `gesso-breakpoint(tablet-lg)`).  The `breakpoint-max` and `breakpoint-min-max`
-mixins can also take an optional parameter to subtrax one pixel from the
-max-width value, which can be useful when using the Gesso breakpoint token
-values.
+mixins can also take an optional parameter to subtract one pixel from the
+max-width value, which can be useful when you want your query to go up to the
+value but not to include it, such as when using Gesso breakpoint token values.
 
 **`@include breakpoint($width) { // styles }`
 Output a min-width based media query.**
--- a/README.md
+++ b/README.md
@@ -165,11 +165,11 @@ examples:
   display: flex;
 }
 
-@include breakpoint-max(gesso-breakpoint(mobile)) {
+@include breakpoint-max(gesso-breakpoint(mobile), true) {
   display: none;
 }
 
-@include breakpoint-min-max(gesso-breakpoint(mobile), gesso-breakpoint(tablet)) {
+@include breakpoint-min-max(gesso-breakpoint(mobile), gesso-breakpoint(tablet), true) {
   display: block;
 }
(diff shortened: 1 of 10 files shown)


kmonahan:
Looks good at first glance! I'll try to test in properly before the next Gesso check-in.

I lean toward defaulting toward subtracting 1px with a flag to disable, because when I'm using max-width queries, I typically mean "up to but not including the specified breakpoint", especially when doing something like:
```
@include breakpoint-minmax(gesso-breakpoint(tablet), gesso-breakpoint(desktop)) {
  // Tablet-only styles
}
@include breakpoint(gesso-breakpoint(desktop)) {
  // Desktop only styles
}
```

But if you and @dcmouyard more typically _do_ want your styles included at the width set as the max-breakpoint, we can set the default the other way.

dcmouyard:
I don’t have a strong preference either way.

My initial assumption is that when I use a variable or pass in a value to a mixin, it’s using that value rather than performing math on it behind the scenes. That said, I kinda see the usefulness of making an exception for max-width media queries.

coreylafferty:
@kmonahan @dcmouyard I've updated the PR to have an optional flag to subtract 1px from the max-width values.  For the moment, I've set the default on this flag to FALSE, so you have to explicitly ask it to subtract 1px if you want that.  In most cases when using a max-width and one of our breakpoint design tokens, you'd probably want this to be TRUE... but I also see Dan's point that it might be confusing for the default to be doing math you might not expect (maybe leading to cases where people pass in a value where they've already subtracted the single pixel, so now it's minus 2) or cases (like in our grid layout) where we're just passing in straight hardcoded pixels values that aren't tied to any design tokens.

If we confirm this approach, I'll update the README to have a new documentation section for the breakpoint mixins.

kmonahan (approved):
Looks good on my end! I'm good with the approach you have here.

README.md
```diff
@@ -305,6 +305,68 @@ value, if available. If there is no fallback value, the token will be omitted
 from the JavaScript objects. In general, if you want to share a value between
 CSS and JavaScript, you should use simple strings or numbers.
 
+### Width Based Media Queries
+
+Gesso uses custom mixins to specify viewport width based media queries:
+`breakpoint`, `breakpoint-max` and `breakpoint-min-max` for min-width,
```
kmonahan: I think a bulleted list might be easier to read here:
* `breakpoint`: min-width queries
* `breakpoint-max`: max-width queries
* `breakpoint-min-max`: queries with both a min and max width

README.md
```diff
@@ -305,6 +305,68 @@ value, if available. If there is no fallback value, the token will be omitted
...
+max-width or a combination of both, respectively. Each mixin takes one or two
+width parameters, which can be a straight value (e.g., 800px, 40em) or a design
+token value called using the `gesso-breakpoint` function (e.g.,
+`gesso-breakpoint(tablet-lg)`).  The `breakpoint-max` and `breakpoint-min-max`
+mixins can also take an optional parameter to subtrax one pixel from the
```
kmonahan: Typo: `subtrax` should be `subtract`

README.md
```diff
@@ -305,6 +305,68 @@ value, if available. If there is no fallback value, the token will be omitted
...
+max-width value, which can be useful when using the Gesso breakpoint token
```
kmonahan: Suggest replacing with `which can be useful when you want your query to go up to the value but not to include it, such as when using the Gesso breakpoint token values`.

kmonahan (commented):
Left a few suggestions for the documentation

PR #465, opened 2021-08-18
Layout full bleed

coreylafferty:
This adds a mixin that can be used on a component inside an l-constrain to force it to go the full width of the viewport. When full bleed components are needed I typically disable the constrain on the content region and add constrains to individual components, but occasionally I've found it useful to do it this way instead.  I've also included a mixin to undo the effect of this mixin. Unlikely to be used frequently, but I have used it when I want a component to go full bleed in some conditions (e.g., on a landing page) but to remain constrained in others (e.g., on a detail page with a sidebar layout).

### DIFF
--- a/source/_patterns/utility-classes/_alignment.scss
+++ b/source/_patterns/utility-classes/_alignment.scss
@@ -24,6 +24,11 @@
   margin-right: auto;
 }
 
+// Breaks something outside its parent l-constrain or other centered container.
+.u-align-full {
+  @include layout-full-bleed();
+}
+
 // Clear floats
 .u-clear-both {
   clear: both;
--- a/source/_patterns/utility-classes/_alignment.scss
+++ b/source/_patterns/utility-classes/_alignment.scss
@@ -24,11 +24,6 @@
   margin-right: auto;
 }
 
-// Breaks something outside its parent l-constrain or other centered container.
-.u-align-full {
-  @include layout-full-bleed();
-}
-
 // Clear floats
 .u-clear-both {
   clear: both;
--- a/README.md
+++ b/README.md
@@ -165,11 +165,11 @@ examples:
   display: flex;
 }
 
-@include breakpoint(gesso-breakpoint(mobile), 'max') {
+@include breakpoint-max(gesso-breakpoint(mobile), true) {
   display: none;
 }
 
-@media (min-width: #{em(gesso-breakpoint(mobile))}) and (max-width: #{em(gesso-breakpoint(tablet))}) {
+@include breakpoint-min-max(gesso-breakpoint(mobile), gesso-breakpoint(tablet), true) {
   display: block;
 }
 ```
@@ -304,6 +304,71 @@ If your token value is a Sass function, the JavaScript will use the fallback
 value, if available. If there is no fallback value, the token will be omitted
 from the JavaScript objects. In general, if you want to share a value between
 CSS and JavaScript, you should use simple strings or numbers.
+
+### Width Based Media Queries
+
+Gesso uses custom mixins to specify viewport width based media queries:
+* `breakpoint`: min-width queries
+* `breakpoint-max`: max-width queries
+* `breakpoint-min-max`: queries with both a min and max width
+
+Each mixin takes one or two
+width parameters, which can be a straight value (e.g., 800px, 40em) or a design
+token value called using the `gesso-breakpoint` function (e.g.,
+`gesso-breakpoint(tablet-lg)`).  The `breakpoint-max` and `breakpoint-min-max`
+mixins can also take an optional parameter to subtract one pixel from the
+max-width value, which can be useful when you want your query to go up to the
+value but not to include it, such as when using Gesso breakpoint token values.
+
+**`@include breakpoint($width) { // styles }`
+Output a min-width based media query.**
+
+examples:
+
+```
+@include breakpoint(gesso-breakpoint(800px)) {
+  display: flex;
+}
+
+@include breakpoint(gesso-breakpoint(desktop)) {
+  display: none;
+}
+```
+
+**`@include breakpoint-max($width, $subtract_1_from_max) { // styles }`
+Output a max-width based media query. The optional $subtract_1_from_max
+parameter will subtract 1px from the width value if set to `true`
+(default: `false`).**
+
+examples:
+
+```
+@include breakpoint-max(900px) {
+  display: block;
+}
+
+@include breakpoint-max(gesso-breakpoint(mobile), true) {
+  display: none;
+}
+```
+
+**`@include breakpoint-min-max($min-width, $max-width, $subtract_1_from_max)
+{ // styles }`
+Output a media query with both a min-width and max-width. The optional
+$subtract_1_from_max parameter will subtract 1px from the max-width value if
+set to `true` (default: `false`).**
+
+examples:
+
+```
+@include breakpoint-min-max(400px, 700px) {
+  display: flex;
+}
+
+@include breakpoint-min-max(gesso-breakpoint(mobile), gesso-breakpoint(tablet), true) {
+  display: block;
+}
+```
 
 ### Creating New Components
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "gesso",
-  "version": "4.3.0",
+  "version": "4.4.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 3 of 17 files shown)


timnolte:
@coreylafferty if this is going to be added to [Gesso WP](https://github.com/forumone/gesso-wp/) it should be noted that the WordPress Block Editor has specific classes it uses with block when an editor wants a block to be full bleed. For Gesso WP the WordPress Core functionality should be used.

coreylafferty:
> 
> 
> @coreylafferty if this is going to be added to [Gesso WP](https://github.com/forumone/gesso-wp/) it should be noted that the WordPress Block Editor has specific classes it uses with block when an editor wants a block to be full bleed. For Gesso WP the WordPress Core functionality should be used.

@timnolte Can you point me towards what those classes are and what CSS they add?  I suspect they do something different than the purpose of this mixin (counteracting what our l-constrain class does) and both would be needed in different situations, but I could be wrong.

timnolte:
> > @coreylafferty if this is going to be added to [Gesso WP](https://github.com/forumone/gesso-wp/) it should be noted that the WordPress Block Editor has specific classes it uses with block when an editor wants a block to be full bleed. For Gesso WP the WordPress Core functionality should be used.
> 
> @timnolte Can you point me towards what those classes are and what CSS they add? I suspect they do something different than the purpose of this mixin (counteracting what our l-constrain class does) and both would be needed in different situations, but I could be wrong.

@coreylafferty here's the relevant WordPress documentation on it: https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-support/#wide-alignment

coreylafferty:
> @coreylafferty here's the relevant WordPress documentation on it: https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-support/#wide-alignment

@timnolte I don't see anything in there that indicates the alignwide is related to what this new mixin is trying to do.  The CSS in the codepen example I see in those docs has nothing to do with going full bleed when locked inside something that's preventing it from doing so.  This is to force a component to break outside of a parent l-constrain (which has a max-width, padding and is horizontally centered) and go edge-to-edge of the browser viewport.  I don't plan on actually adding this mixin to any out-of-the-box components -- just making it available for the rare use cases where its useful.  

If this PR is approved here, I'll reach out to you to confirm I'm not missing anything before merging to Gesso-WP.

timnolte:
> > > > @coreylafferty if this is going to be added to [Gesso WP](https://github.com/forumone/gesso-wp/) it should be noted that the WordPress Block Editor has specific classes it uses with block when an editor wants a block to be full bleed. For Gesso WP the WordPress Core functionality should be used.
> > > 
> > > 
> > > @timnolte Can you point me towards what those classes are and what CSS they add? I suspect they do something different than the purpose of this mixin (counteracting what our l-constrain class does) and both would be needed in different situations, but I could be wrong.
> > 
> > 
> > @coreylafferty here's the relevant WordPress documentation on it: https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-support/#wide-alignment
> 
> @timnolte I don't see anything in there that indicates the alignwide is related to what this new mixin is trying to do. The CSS in the codepen example I see in those docs has nothing to do with going full bleed when locked inside something that's preventing it from doing so. This is to force a component to break outside of a parent l-constrain (which has a max-width, padding and is horizontally centered) and go edge-to-edge of the browser viewport. I don't plan on actually adding this mixin to any out-of-the-box components -- just making it available for the rare use cases where its useful.
> 
> If this PR is approved here, I'll reach out to you to confirm I'm not missing anything before merging to Gesso-WP.

@coreylafferty that would be the `alignfull` class that is added, which would be the intention of what you describe. And to be clear WordPress doesn't automatically apply CSS to accommodate that but it automatically applies the `alignfull` class. So this mixin should work for the WordPress `alignfull` class.

timnolte:
> > > > > @coreylafferty if this is going to be added to [Gesso WP](https://github.com/forumone/gesso-wp/) it should be noted that the WordPress Block Editor has specific classes it uses with block when an editor wants a block to be full bleed. For Gesso WP the WordPress Core functionality should be used.
> > > > 
> > > > 
> > > > @timnolte Can you point me towards what those classes are and what CSS they add? I suspect they do something different than the purpose of this mixin (counteracting what our l-constrain class does) and both would be needed in different situations, but I could be wrong.
> > > 
> > > 
> > > @coreylafferty here's the relevant WordPress documentation on it: https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-support/#wide-alignment
> > 
> > 
> > @timnolte I don't see anything in there that indicates the alignwide is related to what this new mixin is trying to do. The CSS in the codepen example I see in those docs has nothing to do with going full bleed when locked inside something that's preventing it from doing so. This is to force a component to break outside of a parent l-constrain (which has a max-width, padding and is horizontally centered) and go edge-to-edge of the browser viewport. I don't plan on actually adding this mixin to any out-of-the-box components -- just making it available for the rare use cases where its useful.
> > If this PR is approved here, I'll reach out to you to confirm I'm not missing anything before merging to Gesso-WP.
> 
> @coreylafferty that would be the `alignfull` class that is added, which would be the intention of what you describe. And to be clear WordPress doesn't automatically apply CSS to accommodate that but it automatically applies the `alignfull` class. So this mixin should work for the WordPress `alignfull` class.

@coreylafferty this may help to describe it more fully: https://www.billerickson.net/full-and-wide-alignment-in-gutenberg/

coreylafferty:
> > @coreylafferty that would be the `alignfull` class that is added, which would be the intention of what you describe. And to be clear WordPress doesn't automatically apply CSS to accommodate that but it automatically applies the `alignfull` class. So this mixin should work for the WordPress `alignfull` class.
> 
> @coreylafferty this may help to describe it more fully: https://www.billerickson.net/full-and-wide-alignment-in-gutenberg/

Gotcha.  Yeah, we can apply this mixin to the alignfull class.

coreylafferty:
Updated this to apply to the pre-existing full width class and, since we'll be applying to this `alignfull` in WP, I moved this utility class to the alignment partial and renamed it `'l-align-full`.  

@dcmouyard Can I get your thoughts on this reorganization since you wrote the original class?

source/_patterns/utility-classes/_alignment.scss
```diff
@@ -24,6 +24,11 @@
   margin-right: auto;
 }
 
+// Breaks something outside its parent l-constrain or other centered container.
+.u-align-full {
+  @include layout-full-bleed();
+}
```
dcmouyard: I don’t this utility class has anything to do with alignment. I’d prefer if we kept the old `u-full-width` class.

coreylafferty: I don't necessarily disagree. But since WordPress has an internal class called alignfull (to go along with alignleft, alignright, etc), that sort of pushed me to think "Well, it kind of IS about alignment (to the screen anyway) and we already have an alignment file and this would make it easier to be consistent with WordPress' categorization and class names by consolidating." 

I'm a bit torn. 

@timnolte @kmonahan @tomealter or others -- any strong feelings one way or the other? If we kept it as full-width, on the WP side I'd probably apply the mixin to both classes (full-width and alignfull).

kmonahan: I don't know about _strong_ feelings, but I lean toward the second option: keep the utility class name `u-full-width` and also apply the mixin to the WP-specific class in gesso-wp.

timnolte: Well so I guess my hangup is that WordPress also has `alignwide` which should not be full width. It should be wider than the main content area, but not as wide as the full alignment. And I would argue that this is alignment related since even in a desktop text/document editor such as Word making text full width(going from edge-to-edge between the page margins) is an alignment formatting.

IMO I want the theme to be following what is compatible and a standard for the CMS. When we go against the CMS then we can't use out-of-the-box features and it's a maintainability pain. And I don't want to reinvent the wheel because something non-WordPress is dictating how we use the CMS.

timnolte: If we still account for what WordPress needs then I'm fine with the mixin naming so long as we get the right support out-of-the-box from Gesso/PL.

coreylafferty: Ok, I've reverted this to the `u-full-width` class.  I'll create a similar PR to the WordPress side of things that will be the same, except I will also apply the same mixin to the `alignfull` class.

coreylafferty: @dcmouyard @kmonahan I think this is ready to merge in, if you're good with it.  I want to get your thoughts on the addition I made to the `l-site-container` class though.  I realized during testing that the full bleed class creates a horizontal scroll bar (at least in some browsers and OSs).  The article Tim listed above suggests adding `width: 100%; overflow-x: hidden;` to a site wrapper, as did most of the pages I found talking about this full width approach.

kmonahan: There's another approach we used on EPA, where you set the value of a CSS variable with JS to account for the width of the vertical scrollbar (which is what typically causes the overflow with 100vw). Are we trying to keep Gesso IE11 compatible?

tomealter: Added benefit of `overflow-x: hidden;` - I have been using the `l-site-container` component to place max-widths on my last couple of sites for the whole page (including full-bleed components). This is to avoid reeeally wide full-bleed components on very wide monitors. I spun up a quick example in codepen and placing the `overflow-x: hidden;` appears to work like a charm 👍 
https://codepen.io/tomealter/pen/yLXyBoW?editors=1100

coreylafferty: Thanks, @tomealter!

@kmonahan I'd rather not invite JS to the party unless CSS can't handle hosting the festivities alone.  I've used the overflow trick before and rarely run into issues.  Biggest issue might be really wide tables that would normally overflow the site and force a scrollbar.  But now that I think about it, we should probably have a responsive table fix in place out-of-the-box.  I'll create a separate PR for that.

dcmouyard (approved):

coreylafferty:
@dcmouyard I updated the docs here to address I question I think you asked but I originally misunderstood. If the site container wrapper has a max-width, anything using this utility class/mixin will get its edges cropped since it will still go to the edge of the viewport, even those the site container does not.  A rare case, so I added some comments to point out you should not combine those two things.  Can you re-review and let me know if you don't think that's enough to address things?

coreylafferty:
@kmonahan @dcmouyard Can you take another look at this?  I realized after thinking about this a bit that the purpose of the `l-site-container` div is often to add a max-width to all the site content.  If you did that with my previous code AND used the full-bleed class or mixin, the element it is applied to would get cropped (as it would still go to the edge of the viewport, but the site container div would cut off its edges).  

I've moved the `overflow-x: hidden` and `width: 100%` properties to stop the scrollbar from the `l-site-container` div to the `body` tag.  Doing that, we still don't get a scrollbar and anything with full bleed applied will be visible edge to edge of the viewport, no matter if you have a max-width site container or not.

dcmouyard (approved):
The new changes look good to me. I like the clearer documentation.

kmonahan (approved):

PR #467, opened 2021-08-25
adds responsive layout table

coreylafferty:
This code is based off of: https://adrianroselli.com/2020/11/under-engineered-responsive-tables.html

I initially tried adding `display: block` and `overflow-x: auto` to all table elements, which is simpler and works in many cases.  However, by changing the display property of the table element, tables (or more accurately, their child elements) cease to go full width even with `min-width: 100%` on the table element, so I don't think that approach can be applied across the board for all tables.  

After doing some research, it seems a wrapper element around the table is the preferred method for making large tables scrollable at mobile size, so I've added this layout for that purpose.

### DIFF
--- a/source/_patterns/03-layouts/responsive-table/_responsive-table.scss
+++ b/source/_patterns/03-layouts/responsive-table/_responsive-table.scss
@@ -1,0 +1,3 @@
+.l-responsive-table {
+  overflow-x: auto;
+}
--- a/source/_patterns/03-layouts/responsive-table/_responsive-table.scss
+++ b/source/_patterns/03-layouts/responsive-table/_responsive-table.scss
@@ -1,3 +1,4 @@
 .l-responsive-table {
+  @include focus();
   overflow-x: auto;
 }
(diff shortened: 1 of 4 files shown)


source/_patterns/03-layouts/responsive-table/_responsive-table.scss
```diff
@@ -0,0 +1,3 @@
+.l-responsive-table {
+  overflow-x: auto;
+}
```
dcmouyard: We should add focus styles using `@include focus();`.

dcmouyard (changes requested):
Looks good! Just need to fix a typo and add focus styles.

source/_patterns/03-layouts/responsive-table/responsive-table.yml
```diff
@@ -0,0 +1,72 @@
+---
+labeld_by: 'tablecaption01'
```
dcmouyard: Typo: this should be `labeled_by: 'tablecaption01'`

PR #495, opened 2021-11-13
Add Drupal config

### LINKED ISSUE #487: [Gesso 5] Add Drupal config
We should add Drupal config to the theme so that theme settings and block placements are taken care of when the theme is first installed.

I’ve already taken a first stab at this, but I won’t create a PR until we’ve created the 5.x branch in this repo.

dcmouyard:
This PR closes #487.

### DIFF
--- a/config/install/gesso.settings.yml
+++ b/config/install/gesso.settings.yml
@@ -7,7 +7,7 @@ logo:
   use_default: false
 favicon:
   use_default: true
-include_back_to_top: 0
+include_back_to_top: 1
 threshold: '200'
 smooth_scroll: 1
 include_current_page_in_breadcrumb: 1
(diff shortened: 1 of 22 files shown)


config/install/gesso.settings.yml
```diff
@@ -0,0 +1,13 @@
+features:
+  node_user_picture: false
+  comment_user_picture: true
+  comment_user_verification: true
+  favicon: false
+logo:
+  use_default: false
+favicon:
+  use_default: true
+include_back_to_top: 0
```
kmonahan: Wondering if this should default to on rather than off. Do you think people will still know it's there for use if it's off by default?

kmonahan (approved):
Love it! One question about the default Back to Top setting for you to consider -- good to merge once you've decided whether or not to make a change.

config/install/gesso.settings.yml
dcmouyard: Good catch! I meant to match the default set in `theme-settings.php`. I’ll make that change now.

PR #497, opened 2021-11-13
Style default buttons as links

### LINKED ISSUE #488: Default buttons should look like links
Since we only style buttons to look like buttons with the `.button` class, should we style default buttons to look like links instead of the browser default?

dcmouyard:
This PR closes #488.

This work could potentially affect button styling from Drupal core and contrib styles, so I think this needs thorough QA before it gets merged in. I’m going to try this approach on NRDC if we want to use that project as a guinea pig.


source/01-global/normalize/_normalize.scss
```diff
@@ -53,11 +53,22 @@ select {
...
+  border: 0;
+  border-radius: 0;
+  box-shadow: none;
+  cursor: pointer;
+  font-weight: gesso-font-weight(regular);
+  margin: 0;
+  overflow: visible;
+  padding: 0;
+  text-align: left;
+  text-decoration: underline;
```
kmonahan, line 66: I added a text-button mixin that you can use here to make buttons look like regular text links: https://github.com/forumone/gesso_next/blob/main/source/00-config/mixins/_button.scss#L101-L119

dcmouyard: Thanks! I switched to using that mixin to account for a bunch of the default button styles.

kmonahan:
This PR is from 2021 (!!). Reviewing the issue, seems like @dcmouyard , you validated this approach on NRDC and confirmed the buttons it applied to made more sense styled as links rather than matching `c-button`. Confirming this one still needs review + eventual merge?

dcmouyard:
I still haven’t run into any problems with this approach, and I would use it again on all of my future projects.

kmonahan (approved):

PR #502, opened 2021-11-15
Button group

### LINKED ISSUE #491: [Gesso 5] Add new Button Group component
Proposal:
- Add a new Button Group component that displays a group of buttons.
- Create variants of the Button Group to match the different Button variants.
- Add a new Base button variant that can be used for Drupal UI stuff (like tabs.) I’ve called it “Base” for now since there’s a similar variant in USWDS, but is there a better name for it?
- Replace the Tabs component with the new Base Button Group variant

dcmouyard:
This PR fixes #491.


source/03-components/button/_button.scss
```diff
@@ -18,21 +18,28 @@
...
-    color: gesso-color(button, danger, text-active);
-  }
+  @include button(
+    $color-background: gesso-color(button, danger, background),
+    $color-text: gesso-color(button, danger, text),
+    $color-background-hover: gesso-color(button, danger, background-hover),
+    $color-text-hover: gesso-color(button, danger, text-hover),
+    $color-background-active: gesso-color(button, danger, background-active),
+    $color-text-active: gesso-color(button, danger, text-active)
+  );
```
kmonahan: There's a lot of repeated styles here, because button--danger also gets the button class. If your thought is to use the button mixin for all button variants, then it either shouldn't be included on the base `.button` or button variants shouldn't also get the `.button` class. This may also require some updates to the button field formatter to make sure it outputs the right combo of classes for button styles.

source/03-components/button/_button.scss
```diff
@@ -18,21 +18,28 @@
   }
 }
 
+// This button class is used for Drupal tabs.
+.button--base {
+  @include button(
+    $color-background: gesso-color(button, base, background),
```
kmonahan: Is grey what USWDS uses? It seems like an odd default to me, because I typically interpret grey buttons to be disabled.

kmonahan (changes requested):
Only merge-blocking is the repeated CSS rules between button and button--variant. Button group itself looks great.

source/03-components/button/_button.scss
dcmouyard: Yep, USWDS uses gray for their Base button variant. I’m not a fan of the “Base” name, but I’m not sure what to call it.

source/03-components/button/_button.scss
dcmouyard: Hmm, good point. I’ll change the variants to not use the button mixin.

source/03-components/button/_button.scss
dcmouyard: Yep, USWDS uses gray for their Base button variant. I’m not a fan of the “Base” name, but I’m not sure what to call it.

source/03-components/button/_button.scss
dcmouyard: Good point. I changed the variants to not use the button mixin.

kmonahan (approved):

PR #503, opened 2021-11-16
[486] Menu integration

### LINKED ISSUE #486: [Gesso 5] Menu integration issues
Once the menus are complete in Storybook, we need to get them working in Drupal. Current issues noted by @dcmouyard :

- Dropdown menu script is attached to page but doesn't seem to be working
- Mobile menu script is working but hamburger menu isn't styled

kmonahan:
Closes #486

### DIFF
--- a/source/03-components/mega-menu/mega-menu.scss
+++ b/source/03-components/mega-menu/mega-menu.scss
@@ -2,6 +2,16 @@
 // Styles for the Mega Menu.
 @use '00-config' as *;
 
+// Adjust the selector and colors as needed for your site.
+// If the mega menu is not the top element on the page,
+// the element(s) above it need a background color and
+// z-index in order to cover the mega menu.
+.l-header {
+  background: gesso-color('background', 'site');
+  position: relative;
+  z-index: 2;
+}
+
 .mega-menu {
   align-items: center;
   background: gesso-color('background', 'site');
@@ -45,6 +55,7 @@
 
     button.mega-menu__link {
       @include text-button;
+      font-size: rem(gesso-font-size(3));
     }
   }
 
--- a/gesso.libraries.yml
+++ b/gesso.libraries.yml
@@ -6,8 +6,7 @@ global:
 hamburger_button:
   css:
     theme:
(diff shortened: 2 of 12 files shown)


source/03-components/mega-menu/mega-menu.scss
```diff
@@ -2,6 +2,16 @@
...
+// Adjust the selector and colors as needed for your site.
+// If the mega menu is not the top element on the page,
+// the element(s) above it need a background color and
+// z-index in order to cover the mega menu.
+.l-header {
+  background: gesso-color('background', 'site');
+  position: relative;
+  z-index: 2;
+}
+
```
dcmouyard, line 14: I like the documentation here. We should move the `l-header` styles to the header layout partial.

dcmouyard (changes requested):
@KJ great work, as usual!

I think moving the `.l-header` styles to the header layout partial is the only merge blocker.

~~I also noticed that the tabbing doesn’t cycle in the mobile menu in Drupal. It does, however, cycle in Storybook. This issue occurs in the Overlay and Side menus as well. I’m not sure if this should be addressed in this PR or a new one.~~

Just saw #505, so this issue is probably taken care of there.

source/03-components/mega-menu/mega-menu.scss
kmonahan: @dcmouyard I thought about that but intentionally did **not** put the styles there because I only want them to apply when the mega-menu CSS file/mega-menu library is included. I'm assuming a developer will move them around as needed when adjusting the styles for their individual site. I also considered not putting any default styles on the header, but the mega menu looks broken on first enabling otherwise and it's likely not going to be clear to a developer how to fix it. So this seemed like the best compromise to me. Prefer to leave as-is.

dcmouyard: @kmonahan Ah, I had forgotten that this is not a partial that will go into `styles.css` so including it here makes more sense. 👍

dcmouyard (approved):

PR #523, opened 2021-11-29
Accessible Accordion

tomealter:
This is a PR for the accessible accordion and accordion-item storybook components. This accordion component uses aria-labels and is keyboard accessible via the arrow keys.


source/03-components/accordion/accordion.twig
```diff
@@ -0,0 +1,20 @@
+{#
+  "data-allow-multiple" allows multiple accordion items to be open at once.
+  Remove this if you wish to only toggle one at a time.
+#}
```
dcmouyard: I recommend creating a new `allow_multiple` variable that we can use to control this in the Storybook UI and pass as an option in Drupal templates.

source/03-components/accordion/accordion.twig
```diff
@@ -0,0 +1,20 @@
+{#
```
dcmouyard: We need to add `gesso/accordion` as a new library in `gesso.libraries.yml` and add it to the top of this file: `{{ attach_library('gesso/accordion') }}`.

source/03-components/accordion-item/accordion-item.twig
```diff
@@ -0,0 +1,36 @@
...
+
+{% set classes = [
+  'accordion-item',
+  'js-accordion-item',
+  modifier_classes ? modifier_classes : ''
+] %}
+
+<div {{ add_attributes({ class: classes }) }}>
+  <div class="accordion-item__panel expandable">
+    <h3 class="accordion-item__heading">
```
dcmouyard: We should have a variable to set the heading element, and `h3` can be the default.

source/03-components/accordion-item/accordion-item.twig
```diff
@@ -0,0 +1,36 @@
...
+{% set button_id = 'accordion-button'|unique_id %}
+
+{% set classes = [
+  'accordion-item',
+  'js-accordion-item',
+  modifier_classes ? modifier_classes : ''
+] %}
+
+<div {{ add_attributes({ class: classes }) }}>
+  <div class="accordion-item__panel expandable">
```
dcmouyard: Why is the `expandable` class here? I don’t see it getting used anywhere.

source/03-components/accordion-item/accordion-item.twig
```diff
@@ -0,0 +1,36 @@
+{% set section_id = 'accordion-section'|unique_id %}
+{% set button_id = 'accordion-button'|unique_id %}
+
+{% set classes = [
+  'accordion-item',
+  'js-accordion-item',
+  modifier_classes ? modifier_classes : ''
```
dcmouyard: Minor nitpick: all array items should have a trailing comma.

source/03-components/accordion-item/accordion-item.twig
```diff
@@ -0,0 +1,36 @@
...
+        id="{{ button_id }}"
+        aria-expanded="false"
+        aria-controls="{{ section_id }}"
+      >
+        {{ accordion_title }}
+        <span class="accordion-item__icon"></span>
+      </button>
+    </h3>
+    <div
+      class="accordion-item__drawer expandable__drawer"
```
dcmouyard: Why is the `expandable__drawer` class here? I don’t see it getting used anywhere.

source/03-components/accordion-item/accordion-item.twig
```diff
@@ -0,0 +1,36 @@
...
+      id="{{ section_id }}"
+      role="region"
```
dcmouyard: I think we should make `role="region"` optional because having too many `region` landmarks on a page can make it cumbersome to navigate for screen reader users.

Another option would be to count the number of items there are, and only add the `region` landmark if there are less than 5 or so.

source/03-components/accordion-item/accordion-item.twig
```diff
@@ -0,0 +1,36 @@
...
+      aria-labelledby="{{ button_id }}"
+      aria-hidden="true"
+      aria-live="polite"
```
dcmouyard: I was really excited when you mentioned using `aria-live="polite"` in your presentation since I could imagine that being very helpful. After testing it out in a screenreader, though, it seems superfluous and breaks the natural interface of reading through content on a screenreader. It’s not a big deal if there’s very little content in the expanded content area, but it get’s very annoying if there’s more content in there.

Perhaps we could hide this attribute behind a variable that we turn off by default?

source/03-components/accordion-item/_accordion-item.scss
```diff
@@ -0,0 +1,73 @@
...
+}
+
+.accordion-item__drawer {
+  background: gesso-grayscale(white);
+  display: none;
+}
+
+.accordion-item__drawer-inner {
+  padding: 0 rem(gesso-spacing(8)) rem(gesso-spacing(2)) rem(gesso-spacing(2));
+}
```
dcmouyard: Should we add the following to prevent content from adding too much margin at the bottom??

```scss
.accordion-item__drawer-inner {
  > :last-child {
    margin-bottom: 0;
  }
}
```

source/03-components/_index.scss
```diff
@@ -1,4 +1,6 @@
 @use '00-cms-styles/toolbar-menu/toolbar-menu';
+@use 'accordion/accordion';
+@use 'accordion-item/accordion-item';
```
dcmouyard: If we end up having accordion styles be in its own stylesheet rather than in the main one, these can be removed.

source/03-components/accordion/accordion.stories.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
```
dcmouyard: This line isn’t needed since we’re not using `React` in this file.

source/03-components/accordion/accordion.es6.js
```diff
@@ -0,0 +1,170 @@
+import Drupal from 'drupal';
+import { slideDown, slideUp } from '../../06-utility/_slide.es6';
+
+Drupal.behaviors.accordion = {
+  attach(context) {
+    const ACCORDION_CLASS = 'js-accordion';
+    const ACCORDION_TOGGLE_CLASS = 'js-accordion-toggle';
+    const ACCORDION_SPEED = 250;
```
dcmouyard: These variable names should be camel case.

source/03-components/accordion/accordion.es6.js
```diff
@@ -0,0 +1,170 @@
...
+        }
+      });
+
+      // Bind keyboard behaviors on the main accordion container
+      accordion.addEventListener('keydown', event => {
+        const currentTarget = event.target;
+        const key = event.which.toString();
+
+        // 33 = Page Up, 34 = Page Down
+        const ctrlModifier = event.ctrlKey && key.match(/33|34/);
```
dcmouyard: Instead of hard-coding the keycodes, we should pull in constants from `source/00-config/_KEYCODE.es6.js`.

source/03-components/accordion/_accordion.scss
```diff
@@ -0,0 +1,2 @@
+// @file
```
dcmouyard: Minor nitpick: we don’t use `// @file` at the beginning of Sass partials anymore.

source/03-components/accordion/_accordion.scss
```diff
@@ -0,0 +1,2 @@
...
+// Component: Accordion Group
```
dcmouyard: Minor nitpick: I think “Accordion” is a better name.

dcmouyard (changes requested):
@tomealter Great work! I’m glad somebody finally took the time to add an accordion component to Gesso since we use it so often on projects.

You were using an older version of `npm`, so the `package-lock.json` file was inadvertently overwritten in your PR. I merged in the latest `5.x` branch and reverted the `package-lock.json` change.

We need to add accordion as a library in `gesso.libraries.yml`. I also think we should rename `_accordion.scss` to `accordion.scss` and have Drupal only add that stylesheet when needed.

I don’t think we need to have the accordion-item sub-component in a separate folder. I vote to move `accordion-item.twig` into the accordion directory and copy the styles in `_accordion-item.scss` to `accordion/_accordion.scss`. We can then delete everything else in the accordion-item directory. We would also need to update `accordion.stories.jsx`.

A nice-to-have would be to add the option to set an accordion item open by default using a data attribute. This isn’t a merge blocker.

Finally, I think we should update this component to work & look good when JS is disabled. If JS is disabled, we shouldn’t have the button element on the page and the +\- icon shouldn’t be displayed. This isn’t worth blocking the merge, though, so it can be in a follow-up PR.

source/03-components/accordion-item/accordion-item.twig
dcmouyard: 12/2 Meeting recommendation: Perhaps remove region attribute.

source/03-components/accordion-item/accordion-item.twig
dcmouyard: 12/2 meeting notes: perhaps remove this attribute and not use a variable.

source/03-components/accordion/accordion.es6.js
kmonahan: @dcmouyard Was this getting flagged by ESLint or Codacy? If not, I think it's fine as-is. All-caps constant names is a convention I'm familiar with.

source/03-components/accordion/_accordion.scss
kmonahan: @dcmouyard When did we decide that? It's still in the component generator.

source/03-components/accordion/accordion.es6.js
dcmouyard: @kmonahan This wasn’t flagged by ESLint or Codacy, which I’m surprised by. I guess I didn’t know our linting standards as well as I thought! :)

I’m fine with leaving it as-is.

source/03-components/accordion/_accordion.scss
dcmouyard: @kmonahan Hmm, I don’t remember when we made this change, but I believe it was relatively recently.

source/03-components/accordion/accordion.twig
tomealter: Variable added

source/03-components/accordion/accordion.twig
tomealter: Library added and css removed from components index.scss file

source/03-components/accordion-item/accordion-item.twig
tomealter: Heading element variable added

source/03-components/accordion-item/accordion-item.twig
tomealter: Remvoed

source/03-components/accordion-item/accordion-item.twig
tomealter: Trailing commas added

source/03-components/accordion/accordion.stories.jsx
tomealter: Removed

source/03-components/accordion/accordion.es6.js
tomealter: 👍

source/03-components/accordion/_accordion.scss
tomealter: Removed

source/03-components/accordion/_accordion.scss
tomealter: Changed

source/03-components/accordion/accordion.es6.js
tomealter: Replaced key board matches

tomealter:
@dcmouyard and @kmonahan thanks for the notes. I have made the adjustments you recommended.

tomealter:
This recent push adds some markup and js that makes sure the accordions display as expanded in the event that JS does not fire. There is some layout shifting, but it appears to be very quick. I'd like other people's opinions though.

source/03-components/accordion/accordion.twig
dcmouyard: @tomealter @kmonahan The variable works on page load, but the Storybook control doesn’t change the JS behavior. Is there an easy way to get this to work? If not, perhaps we should hide this control in Storybook.

tomealter: @dcmouyard It appears there is some caching going on here or something. If you change your selection and then refresh the page. It works as expected. 🤷

tomealter:
Changed the library settings to load the accordion js in the `<head>` for quicker load time.

dcmouyard:
Got `npm` errors when merging in 5.x. Need to try again and verify which versions of `node` and `npm` we’re using. We should be using `node` v14 and `npm` 8.5.

coreylafferty:
@dcmouyard I've updated this branch with the latest changes on 5.x.  Somehow the package lock file got out of sync, but now both package files should be the same as on 5.x.  I had no issue running `npm ci` and `npm run build`.

dcmouyard:
@tomealter @coreylafferty I updated the accordion component to match what we did on NRDC. Although I approve these changes, it’s probably best that someone else review the changes I made.

tomealter:
@coreylafferty @dcmouyard I made some slight tweaks to the scss, but I think this is good to go otherwise!

dcmouyard (approved):

coreylafferty (approved):

PR #525, opened 2021-12-02
Convert ESLint config to .json files

kmonahan:


kmonahan:
Closing this one. It was being used for some testing only.

Issue #494, opened 2021-11-12
[Gesso 5] Fix display of QuickEdit


dcmouyard:
The QuickEdit UI currently looks broken in Gesso. We need to override some styles to make it look & work correctly.

coreylafferty:
Fixed by #529

PR #529, opened 2021-12-03
Prefix component classes with c-

### LINKED ISSUE #526: Prefix all component classes with `c-`
How do folks feel about prefixing all component classes with `c-` to make it more obvious they’re components? This would also help prevent css class name collisions with Drupal core and contrib styles.

Before:
```scss
.article {}
```

After:
```scss
.c-article {}
```

dcmouyard:
Fixes #526. I didn’t prefix`id` with `c-`, but I can add that to this PR if folks want. (I’m on the fence about that change.)


dcmouyard:
I just confirmed that this PR also fixes #494.

dcmouyard:
@coreylafferty I confirmed that this PR doesn’t break conditional form items.

coreylafferty (approved):

PR #530, opened 2021-12-03
Add shame to wysiwyg.scss

dcmouyard:
This might be my new favorite branch name: `more-shame`


kmonahan (approved):
Good catch and even better branch name :)

PR #533, opened 2021-12-03
Remove instances of route:button

dcmouyard:


source/03-components/mega-menu/_menu-links-macro.twig
```diff
@@ -51,11 +51,8 @@
       } %}
 
       <li {{ add_attributes(additional_item_attributes) }}>
-        {% if menu_item.url == 'route:button' %}
-          <button {{ add_attributes(link_attributes) }}>{{ menu_item.title }}</button>
-        {% else %}
-          {{ link(menu_item.title, menu_item.url, link_attributes) }}
-        {% endif %}
+        {{ link(menu_item.title, menu_item.url, link_attributes) }}
```
kmonahan: @dcmouyard Did you confirm that this still works in Storybook? I think we might need to keep the URL check to ensure we can demo a dropdown menu with button

dcmouyard: Good catch, I’ll confirm.

dcmouyard: @kmonahan The dropdown and mega menus still worked in Storybook, which is why I didn’t notice anything. My changes prevented folks from setting buttons in other menu locations, though, so I fixed that edge case.

kmonahan (approved):

PR #540, opened 2021-12-15
Update node packages

dcmouyard:
This PR is the result of `npm update`. Should we also update the versions in `package.json`?


kmonahan:
I don't think we need to, so long as we've got them as `^X.X.X` or `~X.X.X`.

coreylafferty (approved):

PR #543, opened 2021-12-20
Exclude font files

kmonahan:


source/01-global/01-typography/fonts.scss
```diff
@@ -2,6 +2,16 @@
...
+@font-face {
+  font-family: 'Source Sans Pro';
+  src: url('../../fonts/source-sans-pro/sourcesanspro-regular-webfont.woff2')
+      format('woff2'),
+    url('../../fonts/source-sans-pro/sourcesanspro-regular-webfont.woff')
+      format('woff'),
+    url('../../fonts/source-sans-pro/sourcesanspro-regular-webfont.ttf')
+      format('ttf');
+}
+
```
dcmouyard: @kmonahan Should this be commented out as an example since we don’t have these fonts in the theme? Or should we add these font files to Gesso and remove the call to Google Fonts in `.storybook/preview-head.html`?

dcmouyard (changes requested):
Current PR breaks the build since we don’t have the font files in the theme.

source/01-global/01-typography/fonts.scss
kmonahan: Whoops, that's testing code that's not supposed to be there at all. Thanks for catching! I'll pull that back out.

kmonahan: Removed

coreylafferty (approved):

source/fonts/README.md
```diff
@@ -0,0 +1,6 @@
+Font used in a `@font-face` declaration in will be inlined or copied to
```
coreylafferty: Should this be 'Fonts used in a @font-face declaration will be inlined or copied to` ?

coreylafferty (changes requested):

webpack.common.js
```diff
@@ -93,12 +93,17 @@ module.exports = {
         test: /\.scss$/i,
         exclude: /node_modules/,
         use: [
-          MiniCssExtractPlugin.loader,
+          {
+            loader: MiniCssExtractPlugin.loader,
+            options: {
+              publicPath: '../',
+            },
+          },
```
kmonahan, line 101: @coreylafferty Try adding this to your webpack.common.js in GUS5 and see if it fixes the font path issue

source/fonts/README.md
```diff
@@ -0,0 +1,6 @@
+Fonts used in a `@font-face` declaration in will be inlined or copied to
```
coreylafferty: I think the second use of the word 'in' needs to be removed from this line.

coreylafferty (changes requested):

webpack.common.js
coreylafferty: @kmonahan that did it! Thank you!

coreylafferty (approved):

PR #546, opened 2021-12-23
Responsive typography

### LINKED ISSUE #288: Add responsive heading styles
I’ve been implementing responsive typography on projects for the past few years, and I’m wondering if we should add this functionality to Gesso.

At a minimum this would involve making the heading styles responsive, with perhaps new design tokens to easily set the font-sizes for mobile devices.

There are quite a few approaches to responsive typography, so I think we should hash out which approach we take in this issue before a PR is created.

dcmouyard:
This PR fixes #288 and does the following:

- Add responsive font size tokens that can be used in `clamp()`
- Move Fonts and Text Styles to a new Typography folder in Storybook
- Add new Typographic Scale demo in Storybook which uses the responsive font size tokens
- Add new `responsive-font-size()` mixin which is used for headings
- Responsive technique details: https://css-tricks.com/linearly-scale-font-size-with-css-clamp-based-on-the-viewport/

I think this PR needs a full review with the FED team before it gets merged in since it’s pretty opinionated regarding the responsive typography technique.

### DIFF
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -145,6 +145,10 @@ function getScalarVisitor(path) {
       return identityTransformer;
     }
 
+    if (prefix === 'responsive-font-size') {
+      return identityTransformer;
+    }
+
     if (prefix === 'color') {
       return colorTransformer;
     }
--- a/source/00-config/config.design-tokens.yml
+++ b/source/00-config/config.design-tokens.yml
@@ -246,47 +246,47 @@ gesso:
       display:
         color: text.primary
         font-family: primary.stack
-        font-size: 10
         font-weight: bold
         line-height: tight
+        responsive-font-size: 10
       h1:
         color: text.primary
         font-family: primary.stack
-        font-size: 8
         font-weight: bold
         line-height: tight
+        responsive-font-size: 8
       h2:
         color: text.primary
         font-family: primary.stack
-        font-size: 7
         font-weight: bold
         line-height: tight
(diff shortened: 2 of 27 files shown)


kmonahan:
@dcmouyard If you don't want to use this approach, could you still use breakpoints and the existing `gesso-font-size` functions? That's what it looks like to me (haven't actually pulled it down and tested though) and if so, that makes the opinionated approach here less of a concern. (I'm also fine with Gesso being opinionated, since it's a repository of our best practices, though agree it would be good to get more eyes on this.) Will pull it down and properly review, but I like the general direction here.

dcmouyard:
Labeling this as HOLD since I want to add a few more changes.

dcmouyard:
@coreylafferty @kmonahan I pushed up my latest changes and it’s ready for review. I wasn’t able to figure out how to update `transform.js` to pull responsive font sizes into text styles.

coreylafferty:
@dcmouyard with a lot of help from @kmonahan, I've got the display design tokens working with the new responsive font size tokens.  If you agree with these changes, I think this is ready to merge in.  You can see the changes I made here:

https://github.com/forumone/gesso/pull/546/commits/f3c44e0194e6519ddc5dff4f02d98b4b8bef6651

coreylafferty (approved):

dcmouyard:
@coreylafferty @kmonahan Good job with the responsive text styles!

PR #568, opened 2022-02-01
Fix image path variable

dcmouyard:
This PR fixes the `gesso_image_path` variable so that it works in Storybook both locally and when deployed.


coreylafferty (approved):

coreylafferty:
@dcmouyard @kmonahan I'm not 100% sure why this works both ways (with the ../ and without), but it does so I approved this. If KJ doesn't see an issue with it, I'll merge it in and add to GUS.

kmonahan (approved):

PR #578, opened 2022-02-18
switches to using math.div instead of / in font size mixin

coreylafferty:


timnolte:
Question for my future reference. Is it better to use `math.div` instead of `calc` as the division replacement? As I've been working on security patching updates on Gesso WP on sites, and hitting this linting warning, I had chosen to use `calc` instead of `math.div` I didn't spend time to analyze 1 over the other only that `calc` seems like less code to change and be "easier" to read.

coreylafferty:
@timnolte I just went with the first suggestion the linting error gave (it suggested both options), but after a little googling this is the best answer to your question I could find. @KJ might have other thoughts?

https://stackoverflow.com/questions/70760284/is-there-a-use-case-where-you-would-use-math-div-in-scss-sass-instead-of-a-cal

kmonahan (approved):

kmonahan:
General rule of thumb:
* Use sass math.div() when you want the final CSS to compile to a static value
* Use calc() when you want the final CSS to be a value that the browser will calculate on display

When replacing deprecated sass, math.div() is going to provide parity to the original functionality, while calc() changes it from a single static value to a calculation. So I would say math.div() is the preferable way to replace deprecated /, but the difference is negligible -- other things are going to affect performance far more than some gratuitous calculations and the end result is going to display the same.

PR #593, opened 2022-03-21
Append Sass file to index file when component created

kmonahan:
Adds a `@use` statement to the `_index.scss` file as part of `npm run component` for components that use a sass partial.

### DIFF
--- a/lib/component.js
+++ b/lib/component.js
@@ -254,14 +254,27 @@ async function createComponent(
         library
       )
     );
-    if (
-      !library &&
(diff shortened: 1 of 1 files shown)


coreylafferty:
@kmonahan @dcmouyard What would you think of removing some of the folder choices for where the component lives?  In particular, 00-config, 05-pages, and maybe 06-utility? Those seems like places you wouldn't want to be able to add a full set of component files with Sass, storybook, etc.

kmonahan:
@coreylafferty Let's make that a separate issue instead of discussing here

coreylafferty (approved):

PR #599, opened 2022-05-26
recompile when design tokens change

coreylafferty:


coreylafferty:
@kmonahan Only question I had about this was that '--raw' on the concurrent command.  We already had concurrent running the Drupal task, but it is not on the WP one.  Should they be consistent or is that for StoryBook, so isn't needed in WP?

kmonahan:
@coreylafferty I'll open a PR on the WP repo to add the `--raw` flag. It just turns off output formatting, so it's not a big deal, but makes sense to keep them consistent.

kmonahan (approved):

PR #601, opened 2022-06-02
Add stylelint-order to enforce property order

coreylafferty:


coreylafferty:
@kmonahan is `disableFix: true` line needed in Gesso Drupal?  If so, is it needed in Gesso WP?

kmonahan:
`disableFix: true` turns off Stylelint autofixing for the rule. I wasn't able to find a clear record as to why we added that, but it's been part of Gesso since 3.x. I'm assuming we intentionally didn't want autofixing for this rule and will add to Gesso WP forthwith.

kmonahan (approved):

PR #618, opened 2022-09-21
Fix including common.js JS chunk as Gesso dependency

devbranch-vadym:
Instead of simply defining the `common` library, it's now also added to gesso's `global` library dependencies to make sure it's included everywhere.

### DIFF
--- a/includes/libraries.inc
+++ b/includes/libraries.inc
@@ -6,17 +6,20 @@
  */
 
 /**
- * Implements hook_library_info_build().
+ * Implements hook_library_info_alter().
  */
-function gesso_library_info_build() {
-  $libraries = [];
-  $active_theme = \Drupal::theme()->getActiveTheme()->getPath();
-  if (file_exists($active_theme . '/dist/js/common.js')) {
+function gesso_library_info_alter(&$libraries, $extension) {
+  $activeTheme = \Drupal::theme()->getActiveTheme();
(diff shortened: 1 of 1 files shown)


kmonahan:
With this approach, what prevents a library that depends on gesso/common from being loaded before gesso/global? I tend to lean toward expecting each library to declare its dependencies rather than assuming gesso/global will be on the page, but would love to hear your thoughts.

devbranch-vadym:
> With this approach, what prevents a library that depends on gesso/common from being loaded before gesso/global?

The problem is that you don't know what libraries would depend on `common` unless you build them.
I think `gesso/global` is always included since it's in the theme `info.yml` file. Any library that's in `gesso` theme is expected to be used with the theme only so I think it's safe to assume that `gesso/global` will be always loaded alongside other `gesso` libraries.
As for the load order, chunks are built with Webpack and they could be loaded in any specific order so that's shouldn't be a problem.

We could probably somehow make `webpack` write the chunk dependencies tree into a file (e.g. some sort of debug info as JSON) and conditionally include `common` as dependencies for libraries really needing it; formally that would be the best solution but it sounds like over-engineering :smiley: .

kmonahan:
I think we're safe assuming that any library with a JS file going through webpack depends on common.js -- I agree that we don't need to get into conditionally including common.

PR #619, opened 2022-09-21
Fix theme suggestions for form_element

devbranch-vadym:
This is absolutely non-obvious but theme suggestions having `-` instead of `_` in their names causes Twig templates to not be loaded.

E.g. Twig debug would suggest using `form-element--TYPE.html.twig` template but it doesn't work.

This PR fixes this behaviour.


kmonahan:
Ah, nice, I just ran into that.

kmonahan (approved):

PR #621, opened 2022-09-29
Fix user page templates

dcmouyard:
Individual user pages currently get double titles and local tasks menu. This PR fixes that.


coreylafferty:
@dcmouyard I don't really see any difference with our without these changes in a vanilla Drupal setup. If I apply the page title block to the top of the content area I see two page titles on basic and user pages -- if I disable it, I see one.

dcmouyard:
It turns out this PR isn’t needed until we update how the page title is pulled into the page template.

PR #625, opened 2022-11-04
fix for aria-current not pushing through valid value

coreylafferty:


coreylafferty:
Preference here would be to pass through NULL if not active, but that can cause errors during compilation currently due to issues with add_attributes and twig_drupal_filters. May revert this if #627 is completed or if the underlying packages get updated.

PR #627, opened 2022-11-07
[624] Fix values like false and null getting passed through to front end

### LINKED ISSUE #624: Fork add-attributes-twig-extension
Pull add-attributes-twig-extension JS into Gesso so we can manage it ourselves and update to correctly handle false and NULL values.

kmonahan:
Tracked the issue we were seeing down to the `link()` function, which actually comes from `twig-drupal-filters`, not `add-attributes-twig-extension` like I first thought. [Opened a PR on that repo](https://github.com/kalamuna/twig-drupal-filters/pull/48); the PR here switches to the forked version of the repo with fix until if/when that PR is merged.


coreylafferty:
Converting to draft since this likely needs some additional changes to get the same issue fixed for places where add_attributes is used.  For now, merging in changes from #625 to address this issue (may revert later).

kmonahan:
With the [release of Twig.js 1.16.0](https://github.com/twigjs/twig.js/commit/8d69d0a3d7bb24a790fe0340686bf0f13938a4ec), forked version is no longer needed. 🎉 Still need to fix the issue with add_attributes though.

kmonahan:
> With the [release of Twig.js 1.16.0](https://github.com/twigjs/twig.js/commit/8d69d0a3d7bb24a790fe0340686bf0f13938a4ec), forked version is no longer needed. 🎉 Still need to fix the issue with add_attributes though.

Actually that's not true. https://github.com/kalamuna/twig-drupal-filters/pull/48 is still open

kmonahan:
Forked `add_attributes_twig_extension` (which I don't think is still maintained), pulled it into Gesso, and updated to handle false and null values. Reverted the temporary fix from #625 as part of this work as well.

Note that I also updated the mega_menu Twig file because `aria-current={{ item.is_active ? 'page' : null }}` will never work. Either the entire attribute needs to be in a conditional or add_attributes needs to be used for the attributes. (I went with the second one.) Pretty sure the person who wrote it was definitely thinking of React and not Twig when they did that :disappear:

This is ready for review and applies only to Drupal work, so can be ported to GUS but not Gesso WP or Gesso React.

coreylafferty (approved):

PR #630, opened 2022-11-11
updated packages

coreylafferty:


kmonahan:
Tried to build locally and the compiler exploded -- looks like ESLint errors coming from `import/no-named-as-default` and `import/no-unresolved`. Will see if I can troubleshoot.

kmonahan:
Rolled eslint-import-resolver-typescript back to an earlier version, since it looks like the more recent version requires Node 16, and we're still using Node 14 in docker-compose.gesso.yml. With that change, everything else looks good!

kmonahan (approved):

PR #655, opened 2023-02-09
added front class to homepage in storybook

coreylafferty:


source/05-pages/homepage.stories.jsx
```diff
@@ -59,7 +59,7 @@ const homepageContent = args => twigTemplate({
   homepage_grid_title: 'You Don’t Vote For Kings',
 });
 
-const Homepage = args => <PageWrapper>{parse(homepageContent(args))}</PageWrapper>;
+const Homepage = args => <PageWrapper front="true">{parse(homepageContent(args))}</PageWrapper>;
```
dcmouyard: I think `isHomepage` is a better variable name. Also, we don’t need to add `="true"` here.

Recommended:
```
const Homepage = args => <PageWrapper isHomepage>{parse(homepageContent(args))}</PageWrapper>;
```

source/05-pages/page-wrappers/default.jsx
```diff
@@ -20,7 +20,12 @@ import { BackToTop } from '../../03-components/back-to-top/back-to-top.stories.j
 
 const PageWrapper = props => {
   // eslint-disable-next-line react/prop-types
-  const { children } = props;
+  const { children, front } = props;
```
dcmouyard: We’re also adding the `not-front` class in Drupal, so perhaps we should add it in Storybook as well?

Recommended:
```
  const { children, isHomepage } = props;

  if (isHomepage) {
    document.body.classList.add('front');
    document.body.classList.remove('not-front');
  } else {
    document.body.classList.add('not-front');
    document.body.classList.remove('front');
  }
```

dcmouyard (approved):
I approve of the current changes as-is, but I do have some recommendations.

source/05-pages/homepage.stories.jsx
kmonahan: This should be `front={true}` assuming the intent is to set front as a boolean, not set it to the string "true."

kmonahan: Or just go with Dan's suggestion. I'm good with that too.

PR #666, opened 2023-02-22
Fix duplicate IDs in mobile menu

dcmouyard:


source/03-components/mobile-menu/modules/_MobileMenu.es6.js
```diff
@@ -76,6 +76,14 @@ class MobileMenu extends OverlayMenu {
     if (blockClone.id) {
       blockClone.id = `${blockClone.id}-mobile`;
     }
+    const childrenWithId = [...blockClone.querySelectorAll('[id]')];
+    if (blockClone.matches('[id]')) {
+      childrenWithId.push(blockClone);
+    }
```
kmonahan: I like how you're handling the children, but this seems like it duplicates what's right above it.
```
    if (blockClone.id) {
      blockClone.id = `${blockClone.id}-mobile`;
    }
```
Can lines 80-82 be removed?

dcmouyard: Good catch! Removed redundant code.

dcmouyard:
I added some code to update the `for` attribute of elements to match the updated IDs.

kmonahan (approved):

kmonahan:
@coreylafferty Ready to be merged for the next release. Does not apply to GUS or Gesso WP.

PR #667, opened 2023-02-24
[644] Update dropdown menu when top-level items are buttons

### LINKED ISSUE #644: Update dropdown menu when top-level items are buttons
Currently the dropdown menu has two modes, one where submenu items show on hover and one where submenu items show on click. This is set by whether buttons or links have been created for the top-level items. However, if buttons are used, only one level of submenu items can be seen. There is no way to open a third-level menu.

Proposed solution is to add a button next to child links, so that clicking or touching the button will open the next menu level, similar to how the mobile menu works. Assumption is that the menu should behave consistently, so child links should not open on hover if top-level links require a click.

kmonahan:
Closes #644 . Original Slack discussion that led to that issue at https://forumone.slack.com/archives/CK95A375G/p1668704407523159.


dcmouyard (changes requested):
Great work!

I noticed a bug, however, when tabbing through submenus. If I tab backwards from a link to a button in the previous menu item, the entire dropdown disappears. I expected it to stay open.

kmonahan:
@dcmouyard Hmm, not sure I agree. I think once you're no longer on the submenu, the submenu should hide, same as it does on the mouse-driven version. I could maybe see leaving it open until you activate the button again, but I think if I kept going tabbing backwards, I wouldn't expect the menu to still be open and it seems annoying that if I forgot to close it, I'd have to go back again to close it.

Any example menus you've seen? USWDS looks like it only has one level of submenus in the example.

dcmouyard:
Here’s a video of the bug: 

https://user-images.githubusercontent.com/135259/224133033-ceda6069-3ed3-40b2-84b5-ee23a87c437f.mp4

kmonahan:
Thanks for the demo and video, @dcmouyard ! Pushed an update that should fix that bug.

dcmouyard (changes requested):
@kmonahan I’m still seeing the same issue.

To be extra clear, I was tabbing through the menu items, not using the arrow keys (in case that makes a difference.)

kmonahan:
Iiiiiinteresting. I thought I was also testing with tabs, but I'll see if I'm able to reproduce.

kmonahan:
@dcmouyard Five months later (!!), I was finally able to test and reproduce. Pushed another change that I think should fix it. Give it another try and let me know what results you get.

dcmouyard (approved):
Everything is working as expected for me!

coreylafferty (approved):

coreylafferty:
@kmonahan Looks good! Merging this in. It doesn't look to me like anything here would also apply to WP/React/GUS, but let me know if that's wrong.

PR #670, opened 2023-03-07
Use once() in all relevant JS

tomealter:
Adds core/once to dependencies for mega-menu library and uses once() in mega-menu declaration to avoid loading the js multiple times


dcmouyard:
I updated the other scripts to use `once()`. Fixes #672.

PR #671, opened 2023-03-08
Update unique_id Twig filter

dcmouyard:
This work updates the `unique_id` Twig filter to add a random string rather than using Drupal’s  own `Html::getUniqueId` functionality. This fixes edge cases where caching results in the same ID getting set for multiple items on a page.


kmonahan (approved):
Been using on SLAC for a while, and it's been working nicely 👍

PR #678, opened 2023-03-31
Added additional config for React work; included an example React com…

klatovak:
…ponent.


source/07-react/Example/Example.module.scss
```diff
@@ -0,0 +1,6 @@
+// Component: Example
+
+.blackBox {
+  border: 10px solid #000;
+  padding: 20px;
```
kmonahan, line 6: Let's use our design tokens here, both to bring this example inline without how we'd want people to build, and also to show that you can use those design tokens to style React components just like Drupal ones.

kmonahan (commented):
Apologies @klatovak that it took me roughly 100 years to look at this. Just one small suggested tweak. If you're okay with it, I can make that change and push the changes up to this branch.

kmonahan (approved):
@coreylafferty This is ready to go. Can be ported to GUS but does not apply to Gesso WP or Gesso React

PR #698, opened 2023-05-10
Button data attributes

dcmouyard:
This PR adds a `data_attributes` variable to the Button component, which allows us to manually add data attributes to a button when using the Storybook Twig template directly.

### DIFF
--- a/source/03-components/button/button.twig
+++ b/source/03-components/button/button.twig
@@ -1,3 +1,16 @@
+{% set classes = [
+  'c-button',
+  modifier_classes ? modifier_classes : '',
+]|join(' ')|trim %}
+
+{% set button_attributes = {
+  'class': classes,
+} %}
+
+{% if extra_attributes %}
+  {% set button_attributes = button_attributes|merge(extra_attributes) %}
+{% endif %}
+
 {%- if icon_name and icon_position != 'after' -%}
   {% set button_icon_before %}
     {%- include '@components/icon/icon.twig' with {
@@ -26,11 +39,6 @@
 {% endif -%}
 
 {% apply spaceless %}
-  {% set classes = [
-    'c-button',
-    modifier_classes ? modifier_classes : '',
-  ]|join(' ')|trim %}
-
   {% set button_content %}
     {{ button_icon_before }}
     {{- text -}}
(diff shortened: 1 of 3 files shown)


kmonahan:
I think we should make this more generic and allow for an object containing any extra attributes that will be merged with the class and passed to `add_attributes`. Something like:
```
{% set button_attributes = {
  'class': classes,
} %}
{% if extra_attributes %}
{% set button_attributes = button_attributes|merge(extra_attributes} %}
{% endif %}
<button {{ add_attributes(extra_attributes) ... />
```

Thoughts?

dcmouyard:
> I think we should make this more generic and allow for an object containing any extra attributes that will be merged with the class and passed to `add_attributes`. 

I like that approach and originally considered it, but was worried it’d be overkill. I need to think this through some more to make sure it would be compatible with Drupal’s handling of attributes.

dcmouyard:
I updated this PR to use the object approach KJ suggested and confirmed that it worked with Drupal.

source/03-components/button/button.twig
```diff
@@ -50,13 +58,13 @@
...
+    <button {{ add_attributes(button_attributes) }}{% if is_disabled %} disabled{% endif %}{% if button_type %} type="{{ button_type }}"{% endif %}>{{ button_content }}</button>
   {% else %}
-    <a {{ add_attributes({ 'class': classes }) }} href="{{ url }}"{% if no_follow %} rel="nofollow"{% endif %}{% if storybook %} target="_self"{% endif %}>{{ button_content }}</a>
+    <a {{ add_attributes(button_attributes) }} href="{{ url }}"{% if no_follow %} rel="nofollow"{% endif %}{% if storybook %} target="_self"{% endif %}>{{ button_content }}</a>
   {% endif %}
 
   {% if is_demo %}
-    <a {{ add_attributes({ 'class': classes }) }} href="{{ url }}"{% if no_follow %} rel="nofollow"{% endif %}{% if storybook %} target="_self"{% endif %}>{{ demo_button_content }}</a>
-    <button {{ add_attributes({ 'class': classes }) }} disabled>{{ disabled_button_content }}</button>
+    <a {{ add_attributes(button_attributes) }} href="{{ url }}"{% if no_follow %} rel="nofollow"{% endif %}{% if data_attributes %} {{ data_attributes }}{% endif %}{% if storybook %} target="_self"{% endif %}>{{ demo_button_content }}</a>
```
coreylafferty: Is this data_attributes bit left over from a previous approach?

source/03-components/button/button.yml
```diff
@@ -4,6 +4,7 @@ text_demo: 'Link Button'
 is_demo: true
 is_button_tag: true
 is_disabled: false
+data_attributes: ''
```
coreylafferty: Is this data_attributes bit left over from a previous approach?

coreylafferty (changes requested):

source/03-components/button/button.twig
dcmouyard: Yep, just removed it.

source/03-components/button/button.yml
dcmouyard: Yep, just removed it here too.

coreylafferty (approved):

PR #702, opened 2023-05-19
Danger button

dcmouyard:
This PR fixes the button classes for `delete` form actions and fixes form input attributes.


dcmouyard:
The best place to test this is on a node edit page.

Before:
before

After:
after

coreylafferty (approved):

PR #709, opened 2023-06-20
Add u-visually-hidden class if label_display is invisible

dcmouyard:
We need to add the `u-visually-hidden` class to the form item label if the `label_display` is set to invisible.

### DIFF
--- a/source/03-components/form-item/_form-item-label.twig
+++ b/source/03-components/form-item/_form-item-label.twig
@@ -6,6 +6,7 @@
   'c-form-item__label',
   is_disabled ? 'is-disabled' : '',
   is_required ? 'is-required' : '',
+  label_display == 'invisible' ? 'u-visually-hidden' : '',
   title_display == 'invisible' ? 'u-visually-hidden' : '',
   modifier_classes ? modifier_classes : '',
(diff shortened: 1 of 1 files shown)


coreylafferty:
@dcmouyard Is this necessary?  Looking at this template (and those in core), it looks like `title_display` is the attribute name in this file (which we already have).  `label_display` appears in `form-element.html.twig`, not `form-element-label.html.twig`.  Or am I missing something?

dcmouyard:
@coreylafferty This PR isn’t necessary to make invisible labels work in Drupal, but it’s necessary to make them work in Storybook. For example, this form item label should be invisible: https://forumone.github.io/gesso/?path=/story/components-form-item-input--text&args=label_display:invisible

This is due to the `*.stories.jsx` files pulling in labels with the same data.
```
const label = args => labelTemplate(args);
```

Since Drupal uses two different variable names in these two templates for the same thing, we need to target both variable names in the Storybook label template to get the Storybook control working.

Another fix that would do the same thing would be to do the following instead:
```
{% if label_display %}
  {% set title_display = label_display %}
{% endif %}
```

PR #712, opened 2023-06-30
711: CKEditor 5 font include

### LINKED ISSUE #711: Modals with CKEditor 5 WYSIWYG causes JS error
https://github.com/forumone/gesso/pull/705

The above code seems to break WYSIWYGs in a modal. This is the error in console when the modal fails:

```
webform:1 Refused to execute script from 'https://fonts.googleapis.com/css?family=Crimson+Text:400,400i,600,600i%7CMontserrat:300,400,600,700&display=swap' because its MIME type ('text/css') is not executable, and strict MIME type checking is enabled.
ajax.js?v=10.1.0:1118 An error occurred during the execution of the Ajax response: The following files could not be loaded: //fonts.googleapis.com/css?family=Crimson+Text:400,400i,600,600i|Montserrat:300,400,600,700&amp;display=swap1
(anonymous) @ ajax.js?v=10.1.0:1118
```

Repro steps:
- Install Drupal, Webform, and Gesso
- Attempt to add a webform using the button on /admin/structure/webforms
- Modal should fail
- remove the above change and attempt the webform again
- modal should work

Something seems to be broken with how CKE5 is pulling in any of the resources within that configuration. Or there's something we're not doing correctly there? We might need to include the Google font resource within the wysiwyg css instead of separately, but not sure about how we might do that properly.

Noting here for documentation at any rate.

jbwagner:
Fixes #711

### DIFF
--- a/gesso.info.yml
+++ b/gesso.info.yml
@@ -12,7 +12,6 @@ ckeditor_stylesheets:
   - //fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,300;0,400;0,600;0,700;1,300;1,400;1,600;1,700&display=swap
   - dist/css/ckeditor4-styles.css
 ckeditor5-stylesheets:
-  - //fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,300;0,400;0,600;0,700;1,300;1,400;1,600;1,700&display=swap
   - dist/css/editor-styles.css
 dependencies:
   - components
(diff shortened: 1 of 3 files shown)


gesso.info.yml
```diff
@@ -12,7 +12,6 @@ ckeditor_stylesheets:
   - //fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,300;0,400;0,600;0,700;1,300;1,400;1,600;1,700&display=swap
```
dcmouyard: I think we should move these fonts to the CKEditor 4 stylesheet as well.

dcmouyard (changes requested):

PR #720, opened 2023-07-13
Add default responsive table layout to tables added via WYSIWYG

dcmouyard:


coreylafferty (approved):

source/03-components/wysiwyg/wysiwyg.es6.js
```diff
@@ -0,0 +1,31 @@
...
+
+      // Wrap all tables output via WYSIWYG with a responsive table layout.
+      const tables = wysiwyg.querySelectorAll('[data-wysiwyg] table');
+      tables.forEach((table, i) => {
+        const responsiveTable = document.createElement('div');
+        const caption = table.querySelector('caption');
+
+        responsiveTable.classList.add('l-responsive-table');
+        responsiveTable.setAttribute('tabindex', '0');
+        responsiveTable.setAttribute('role', 'region');
```
coreylafferty: @dcmouyard as I was working on this to port it to GUSWDS, I realized it's possible that someone could have already manually added the responsive layout div around their table and this would add a second one. Do you think it needs to a check to make sure it's direct parent isn't already a responsive table wrapper div?

coreylafferty (changes requested):

source/03-components/wysiwyg/wysiwyg.es6.js
dcmouyard: @coreylafferty Good catch! That edge case should be handled properly now.

coreylafferty (approved):

PR #726, opened 2023-07-19
adds code to remove no-js class when JS present

coreylafferty:
It looks like we left the 'no-js' class on the HTML tag, but no longer have a custom version of modernizr that removes it.  Drupal has a file (core/misc/drupal.init.js) that adds a 'js' class to the HTML tag when JS is present, but it does not take off the 'no-js' class. 

We could just remove our no-js class, but I think there are cases where it could be useful (like when you want to hide something until JS kicks in).  

I've added a new script add the 'js' class and remove the 'no-js' class (which doesn't conflict with what Drupal's doing if it gets there first).

### DIFF
--- a/source/01-global/global.es6.js
+++ b/source/01-global/global.es6.js
@@ -1,5 +1,5 @@
 import Drupal from 'drupal';
-import setJSClass from '../../../06-utility/_js.es6';
+import setJSClass from '../06-utility/_js.es6';
 
 Drupal.behaviors.setJSCLass = {
   attach() {
--- a/source/01-global/global.es6.js
+++ b/source/01-global/global.es6.js
@@ -1,4 +1,4 @@
 if (document.documentElement.classList.contains('no-js')) {
-  document.documentElement.classList.add('js');
   document.documentElement.classList.remove('no-js');
 }
+document.documentElement.classList.add('js');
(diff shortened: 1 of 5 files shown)


source/01-global/global.es6.js
```diff
@@ -0,0 +1,4 @@
+if (document.documentElement.classList.contains('no-js')) {
+  document.documentElement.classList.remove('no-js');
+}
```
dcmouyard: We don’t need to check if it already has the class `no-js` here, the browser will do that automatically with `classList.remove()`.

dcmouyard (changes requested):
We did something very similar on NRDC, but we followed KJ’s example of how they added the scrollbar-width custom property. See #731 for code details.

Basically, we added `source/01-global/html-elements/00-universal/universal.es6.js` and imported the no-js code from its own file at `source/06-utility/_no-js.es6.js`.

I think we should have a consistent way to add global scripts, so which approach should we standardize on?

coreylafferty:
@dcmouyard I updated this with some of your suggestions. I feel like placing the file in the 'universal' folder makes sense for the use case you mention, but as a generic approach I think that buries the file too deep and not all code added here will really be about HTML elements or relate to 'universal'.  Instead I've left the main file in global, but did split the actual class modification code off into a utility file.  What do you think?

PR #739, opened 2023-08-12
Webpack improvements

### LINKED ISSUE #650: Update webpack-cli to 5.x
We're still using 4.x version of webpack-cli, so we should update that to 5.x. There are some breaking changes with the release, but nothing that at first look seems to affect us: https://github.com/webpack/webpack-cli/releases/tag/webpack-cli%405.0.0

kmonahan:
The first couple of fixes here are aimed at speeding up webpack a bit. I backported these from another project, where one of the devs was initially seeing painful startup times. The change won't be dramatic, but I fixed an issue where webpack was going through all the JS and SCSS files twice on start up. (It does have to go through the design tokens twice, but that's pretty quick.) Also updated how the JS and SCSS files in the theme are found by webpack, so it's now asynchronous with a shared cache.

Second set of changes here doesn't have any direct impact on build time but aims at making it clearer what webpack is doing. I turned off a lot of the default output so it's less noisy. (Storybook still prints a lot of unnecessary logs, but that's not an issue if you're using the standard ddev setup.) In it's place, I also added messages indicating when webpack is beginning compliation and when it's done (with a timestamp, so you can tell one update from the next). This is to make it easier to tell if a thing isn't working because there's a bug in it or if a thing isn't working because Webpack hasn't finished building the updated files yet.


webpack.common.js
```diff
@@ -175,5 +179,9 @@ module.exports = {
   },
   output: {
     path: path.resolve(__dirname, 'dist'),
+    clean: {
+      keep: /images/,
+    },
```
coreylafferty: Do we need this? While currently it works with Gesso out of the box, we also allow adding font files to the gitignore (which would be wiped out by this) and I sometimes have /js/library files I add to the gitignore. We can add a comment to the gitignore to be sure to add anything you want to keep to this file as well, but want to make sure we need this before suggesting we add that.

coreylafferty (changes requested):

webpack.common.js
kmonahan: Hmm, good question. Normally I would say yes, we need it, to ensure that webpack doesn't keep generating new versions of the files and keeping the old ones forever. But since we're not including a cache-busting hash in any of our JS or CSS files, we might be okay without it. I'll turn it off for now, and we can add it back if we start seeing old files lingering unnecessarily.

kmonahan: Updated!

coreylafferty (approved):

PR #742, opened 2023-08-19
Import global JS files from source instead of dist

kmonahan:
Fixes `__webpack_require__.[something]` is not a function errors seen in Storybook production builds when JS is not minimized. (Honestly not entirely sure why it was working when JS _is_ minimized, but this should take care of it either way 🤷 )


coreylafferty:
Doh!  Why didn't I think of pulling in the source files? Thank you for the quick turn around!

coreylafferty (approved):

PR #746, opened 2023-09-21
Update html-react-parser to latest version

kmonahan:
The issue with `parse()` stripping the `value` attribute from `<option>` elements just came up again on another project. The [original issue](https://github.com/remarkablemark/html-react-parser/issues/625) was fixed a while ago, but Gesso still relies on a pretty old version. This PR bumps us up to the latest.

Applies to both Gesso and GUSWDS and to Gesso for React (although Gesso for React is on a more recent version then the Drupal editions).


dcmouyard (approved):
👍 Everything compiled correctly for me.

coreylafferty (approved):

PR #747, opened 2023-09-22
Adds webp to the file types used for background images

kmonahan:
Also applies to GUSWDS and Gesso WP. Not needed for Gesso for React, since Next.js handles that webpack config.


dcmouyard (approved):
Change looks good to me.

coreylafferty (approved):

PR #749, opened 2023-10-09
[649] Update YAML

### LINKED ISSUE #649: Update YAML to 2.x
The design tokens functionality currently relies on yaml 1.10.2 to convert the YAML design tokens. We should update to the latest release, 2.2.1, but it looks like there are some breaking changes that will affect us (removal of parseMap from yaml/util) so will require some work to implement.

kmonahan:
Updates `yaml` to 2.x and updates the design tokens processing to work with the latest version. Includes fixes for linting and type-checking. In theory, everything should look the same if you're firing up `ddev gesso watch` or `ddev gesso build` -- changes are all under the hood.

This applies to both Gesso WP and GUS, but since GUS also handles a `uswds` tag, I suspect it'll need some additional work. I'll open a PR for that one in the next coupld weeks.

Closes #649


kmonahan:
Opened https://github.com/forumone/gesso-uswds/pull/264 for GUS

coreylafferty (approved):

PR #750, opened 2023-10-11
refactors demo content for Storybook

coreylafferty:
This refactors all demo content across components, removing anything from the Twig files to do with Storybook or demonstration content and moves it into the Story files and/or YAML, primarily through custom decorators or the new content-placeholder component.


coreylafferty:
@dcmouyard @kmonahan Made some additional tweaks. I think this is ready for review now.

One issue I could use help on is that I'm seeing this eslint error during compile and I'm not sure the best way to fix it.
```
themes/gesso/source/01-global/content-placeholder/content-placeholder.js
error  'children' is missing in props validation  react/prop-types
```

kmonahan:
> @dcmouyard @kmonahan Made some additional tweaks. I think this is ready for review now.
> 
> One issue I could use help on is that I'm seeing this eslint error during compile and I'm not sure the best way to fix it.
> 
> ```
> themes/gesso/source/01-global/content-placeholder/content-placeholder.js
> error  'children' is missing in props validation  react/prop-types
> ```

I think we need to update the eslint config to turn off that rule. It comes from https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/prop-types.md, but since we're neither requiring TypeScript in Gesso for Drupal nor using React outside of Storybook in most cases, it seems less helpful here than it is in, say, Gesso for React.

source/01-global/content-placeholder/content-placeholder.js
```diff
@@ -0,0 +1,18 @@
+// Used for to show simple demo content placeholders within style guide.
```
kmonahan: This should have a `.jsx` extension since it exports a JSX/React component

source/02-layouts/breadcrumb/breadcrumb.stories.jsx
```diff
@@ -1,17 +1,22 @@
...
-      },
-    },
-  },
+  decorators: [
+    (Story, {args}) =>
+      <Story args={{
+        ...args,
+        breadcrumb_content: ReactDOMServer.renderToStaticMarkup(
+          <ContentPlaceholder>Breadcrumb Layout Content</ContentPlaceholder>
+        )
```
kmonahan: This will forcibly override anything you put as `breadcrumb_content` in the args with placeholder content. What it seems like you're trying to do is provide `<ContentPlaceholder>` as the default arg for breadcrumb_content. In that case, I don't think a decorator is the right approach. I think instead you need to take `breadcrumb_content: ''` out of the YAML file and set Breadcrumb.args as 
```
Breadcrumb.args = {
  breadcrumb_content: ReactDOMServer.renderToStaticMarkup(
      <ContentPlaceholder>Breadcrumb Layout Content</ContentPlaceholder>
   ),
  ...data,
}
```
The ordering there is intentional so that if I do put content in breadcrumb_content in the YAML file, that content rather than the placeholder will be used.

You can still override the content anywhere you want to reuse the story by doing
```
Breadcrumb({
   ...Breadcrumb.args,
   breadcrumb_content: ReactDOMServer.renderToStaticMarkup(<div>Some different content here</div>)
})
```
Decorators are for wrapping a story in extra markup or functionality, like we were looking at with the buttons. 

This applies not just here but everywhere this pattern is used.

source/03-components/icon/icon.stories.jsx
```diff
@@ -7,7 +7,18 @@ import data from './icon.yml';
...
+      const icons = globalData.icons.map(icon => (
+        <p key={icon}>
+          <Story args={{ ...args, icon_name: icon }} />
+          &nbsp;&nbsp;{ icon }
+        </p>
+      ));
+      return <>{icons}</>;
+    },
+    withGlobalWrapper,
+  ],
```
kmonahan, line 18: This can be shortened a bit to
```
    (Story, { args }) => globalData.icons.map(icon => (
        <p key={icon}>
          <Story args={{ ...args, icon_name: icon }} />
          &nbsp;&nbsp;{ icon }
        </p>
      ));
    },
```

I do think this one makes sense with a decorator.

kmonahan (changes requested):

source/01-global/content-placeholder/content-placeholder.js
coreylafferty: Making that changed fixed the ESLint error. Thanks!

source/02-layouts/breadcrumb/breadcrumb.stories.jsx
coreylafferty: That makes a lot of sense -- thank you! 

I've updated all the Layout and Template stories to match this updated pattern.  I left the component ones alone (back-to-top, button, icon, icon-link) because they all seemed more like decorators to me, but please let me know if you think additional changes are needed there or elsewhere.

source/03-components/icon/icon.stories.jsx
coreylafferty: Updated.

coreylafferty:
> I think we need to update the eslint config to turn off that rule. It comes from https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/prop-types.md, but since we're neither requiring TypeScript in Gesso for Drupal nor using React outside of Storybook in most cases, it seems less helpful here than it is in, say, Gesso for React.

Changing the filename from JS to JSX seems to stopped this error message from showing. Not sure if we still need to remove that rule, but in terms of this PR its no longer an issue.

kmonahan (approved):

PR #751, opened 2023-10-11
adds focus for radio items

coreylafferty:
Ran into this on a project. Radio items can individually be navigated with up/down arrow keys, but when tabbing through a form there is no focus state to indicate that you're inside the list of radio items. This fixes that, adding a focus state around the first item if none is selected when you tab to them.

### DIFF
--- a/source/03-components/form-item/form-item--radio/_form-item--radio.scss
+++ b/source/03-components/form-item/form-item--radio/_form-item--radio.scss
@@ -9,7 +9,6 @@
   }
 
   .c-form-item__radio {
-    @include focus();
     appearance: none;
     background-color: gesso-color(form, background-unchecked);
     border: 1px solid gesso-color(form, border-dark);
@@ -30,6 +29,10 @@
       background-color: gesso-color(form, background-checked);
       border: 2px solid gesso-color(form, background-unchecked);
       box-shadow: 0 0 0 2px gesso-color(form, background-checked);
+    }
+
+    &:focus-visible {
+      @include focus();
     }
 
     &:disabled {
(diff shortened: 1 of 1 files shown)


source/03-components/form-item/form-item--radio/_form-item--radio.scss
```diff
@@ -32,6 +31,10 @@
       box-shadow: 0 0 0 2px gesso-color(form, background-checked);
     }
 
+    &:focus-visible {
+      @include focus();
```
kmonahan: I'm confused here because the focus mixin already contains styling for the focus state. So what this will compile to is
```
.c-form-item__radio:focus-visible {
  outline: 2px solid transparent;
  outline-offset: 2px;
}

.c-form-item__radio:focus-visible:focus {
  outline-color: gesso-color(ui, generic, focus);
}

.c-form-item__radio:focus-visible:focus:not(:focus-visible) {
  outline-color: transparent
}
```
That seems a bit odd, especially the last part. Am I misunderstanding what's happening here? It seems to me like the focus mixin was in the correct spot originally but that some thing else needs to be added to fix the issue you noted.

kmonahan (changes requested):

source/03-components/form-item/form-item--radio/_form-item--radio.scss
coreylafferty: @kmonahan While my fix did technically work, I think you're right -- it's adding some CSS that doesn't make much sense and there's a simpler solution. I did some additional testing and realized it's the `outline: 0` that was causing the issue (no focus state using keyboard on radios), so I've updated the PR to just remove that.  Everything seems to work now.  To test, just open 'Form Item / Radios' in StoryBook and try to tab to them and get any of the radios to show a focus state without using your mouse.

kmonahan (approved):

PR #752, opened 2023-10-12
adds base font properties from the HTML tag to ckeditor5 content

coreylafferty:
Since CKEditor5 is not in an iframe, any styles applied directly to the BODY or HTML tag do not carry through to the WYSIWYG since those tags don't exist inside the editor. Other element's CSS (headings, lists, etc.) work fine, but the base font properties we typically apply to the HTML tag do not get applied to things like a plain `<p>` tag. This PR addresses that by applying the basic font properties directly to the editor container. 

If this seems hacky, a cleaner approach might be to use a placeholder or mixin... but it seemed like overkill as I started to go down that path.  I'm open to suggestion though if anyone has a more elegant suggestion for how to do this or thinks a mixin or something would be better.

### DIFF
--- a/source/editor-styles.scss
+++ b/source/editor-styles.scss
@@ -8,6 +8,15 @@
 // stylelint-disable selector-class-pattern
 div.paragraph--view-mode--preview,
 div.ck-content {
+  // Ensuring font properties from source/01-global/html-elements/01-html/_html.scss
+  // get applied to CKEditor content. This may need to updated to match any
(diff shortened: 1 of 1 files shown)


kmonahan:
Hmm, that's an excellent question. I initially learned toward the mixin approach, but looked into it a bit more and quickly reached the same opinion as you -- it feels like overkill since we've already got design tokens in place to set the primary font size and line height. I'm good with this as-is.

kmonahan (approved):

PR #765, opened 2023-11-03
[662] Update stylelint LVHFA plugin

### LINKED ISSUE #662: Stylelint gives false error on LVHFA order
When linting for the LVHFA order, Stylelint can throw an error but nothing is technically an issue when looking at the code yourself.

Returns error:
```
a[href]:focus {
  outline-offset: -0.25rem;
}

&:hover,
&:focus {
  img {
    transform: scale(1.1);
  }
}
```

Obviously those are two separate order sets, but Stylelint seems to count them as the same level, failing because focus comes before the first hover. However, you can workaround this by doing the following:
```
a[href] {
  &:focus {
    outline-offset: -0.25rem;
  }
}

&:hover,
&:focus {
  img {
    transform: scale(1.1);
  }
}
```

The above passes linting just fine. I haven't tested specifically, but I would guess dropping the `a` selector in the original example below the pseudo selector set would make it pass as well.

I'm not sure if this is an explicit bug, or theres some other secondary ordering that it's expecting, but notating here to keep track and mention the workaround.

kmonahan:
Fixes #662 and updates the plugin to avoid depending on stylelint internal utils (per their documentation) and to include `:focus-visible` and `:focus-within` in the order. (I'm not sure what is the correct ordering for the F selectors; right now I have `:focus`, `:focus-visible`, `:focus-within`.)


dcmouyard:
> Fixes #662 and updates the plugin to avoid depending on stylelint internal utils (per their documentation) and to include `:focus-visible` and `:focus-within` in the order. (I'm not sure what is the correct ordering for the F selectors; right now I have `:focus`, `:focus-visible`, `:focus-within`.)

I like the `:focus, :focus-visible, :focus-within` order.

dcmouyard (changes requested):
@kmonahan Thanks for addressing this! I noticed two linting errors during testing. I fixed one, but the other I believe is a false positve:

```scss
&:focus:not(:focus-visible) {
  outline-color: transparent;
}
```

Could the plugin be updated to address this example?

kmonahan:
> @kmonahan Thanks for addressing this! I noticed two linting errors during testing. I fixed one, but the other I believe is a false positve:
> 
> ```css-scss
> &:focus:not(:focus-visible) {
>   outline-color: transparent;
> }
> ```
> 
> Could the plugin be updated to address this example?

@dcmouyard I'll take a look but that might be an instance where you need to disable the plugin. I'm not sure we can make that distinction.

kmonahan:
@dcmouyard `&:focus:not(:focus-visible)` was failing because `:focus` was listed twice in that file. Updated the plugin so that duplicates are fine as long as they match the previous selector, so
```
&:focus {
  // Rules
}

&:focus:not(:focus-visible) {
  // Rules
}
```
is fine, but
```
&:focus {
  // Rules
}

&:focus-visible {
  // Rules
}

&:focus:not(:focus-visible) {
  // Rules
}
```
is not.

However, `:focus:not(:hover)` would have still failed, because the plugin sees the ordering as `focus, hover`. I can't entirely handle this scenario, so I excluded `:not()` from linting. Which means that `:focus:not(:hover, :link)` will also pass linting, but I think that's fine, since the ordering within a `:not()` selector doesn't impact the cascade.

dcmouyard (approved):
@kmonahan Works great!

PR #771, opened 2023-11-06
Fix views view title

dcmouyard:
This PR adds a `views_title` variable in the views-view template. By default, Drupal only outputs the `title` variable on admin pages.


includes/views.inc
```diff
@@ -9,7 +9,13 @@
  * Implements hook_preprocess_HOOK().
  */
 function gesso_preprocess_views_view(array &$variables) {
-  $variables['path'] = $variables['view']->getRequest()->getPathInfo();
+  /** @var \Drupal\views\ViewExecutable $view */
+  $view = $variables['view'];
+
+  $variables['path'] = $view->getRequest()->getPathInfo();
+  $variables['view_title'] = [
```
kmonahan: Two things here: 1) I think this should be `$variables['title']` instead of `$variables['view_title']` so it matches Drupal's core `views-view.html.twig` file. This also makes the mapping in the template file optional (although I generally prefer explicit mappings anyway so it's clear what variables are expected). 2) I think it would be good to add a check here first so that we're only getting the title if we don't already have it, either by checking `$variables['view']` or by checking the view display mode.

kmonahan (changes requested):
Couple comments here, but very excited to see this added to Gesso, as this is definitely something I end up doing on EVERY. SINGLE. PROJECT.

includes/views.inc
dcmouyard: @kmonahan Good ideas! I added them to the latest commit.

kmonahan (approved):

coreylafferty:
@kmonahan @dcmouyard I made a couple of updates to this. 

1. Updated the docs to not say 'title only shows in admin view'.
2. Added a customizable H2 wrapper around the title in the view template
3. Added a variable to hide the title, if desired

Let me know if these look good to you? If so I'll go ahead and merge.

kmonahan:
I'm good with these changes. Defer to @dcmouyard though, since he did the original work.

PR #773, opened 2023-11-08
Container query mixins

dcmouyard:
This PR adds container query mixins using the same approach as the breakpoint mixins and adds information about these new mixins to the README.


source/00-config/mixins/_container-query.scss
```diff
@@ -0,0 +1,94 @@
...
+
+// Create a min-width container query.
+// @param {Number} $breakpoint - width value.
+// @param {String} $container-name - name of container.
+@mixin container-query-min($breakpoint, $container-name: null) {
+  @if $container-queries-rems {
+    $breakpoint: rem($breakpoint);
+  }
+  @if $container-name {
+    // stylelint-disable
```
kmonahan: I don't think we should be disabling the linter without comment (even though I see it's re-enabled a few lines later). I'm guessing this is because Stylelint doesn't understand something here? First recommendation would be to update Stylelint as part of this PR. Stylelint _should_ support `@container` now (their source code suggests they do), so Gesso might be running an old version. (Let me know if you need a hand with this.)

If even updating doesn't help Stylelint, we should [follow our own guidelines](https://github.com/forumone/gesso/blob/5.x/README.md#sass-linting) and include a comment about why the disabling of linting is necessary, so that we can (hopefully) remove that once Stylelint catches up.

kmonahan (changes requested):

source/00-config/mixins/_container-query.scss
dcmouyard: @kmonahan I was able to fix the linting issues in the latest commit.

kmonahan (approved):

PR #774, opened 2023-11-09
Add new font scale and accessibility documentation for the responsive font size design tokens

dcmouyard:


kmonahan (approved):
Great idea!

PR #779, opened 2023-12-11
Icon link formatter

kmonahan:
Adds a new field formatter for icon links and modifies the existing button field formatter to allow for icons. Also adds a modifier classes field to both so that additional theme or component-specific classes can be added.


kmonahan:
Drupal-specific: applies to GUS but not to any other Gessos

coreylafferty:
@kmonahan This now has conflicts since the upgrades to all the story files have been merged. I started to fix things, but it got a bit complicated so I think I'll let you resolve these conflicts before I begin testing if that's ok.

kmonahan:
@coreylafferty I won't be able to get to this until March, but I don't think there's a problem with letting it hang out for a bit.

coreylafferty (changes requested):
Currently has conflicts with 5.x/5.x-RC. Please fix conflicts and then will retest.

kmonahan:
@coreylafferty Updated to resolve merge conflicts

dcmouyard (approved):
Works like a charm!

coreylafferty (approved):

PR #781, opened 2024-01-12
updated stories to not show unpublished by default and always load assets

coreylafferty:
@dcmouyard @kmonahan As I was testing this I noticed that:

- In Storybook, the unpublished message displayed everywhere by default
- Unless you'd already clicked on a story within Storybook that loaded the message CSS, the unpublished message was unstyled on templates/pages. This is actually true of the admin controls as well, even in the current release.

To address these issues I've: 

1. Set a global variable to set things in Storybook to published (similar to how we've hidden the admin controls) so you have to manually click to show the unpublished indicator. 
2. Refactored how we're loading things in the templates/pages, loading any necessary stories in templates that might contain styling or JS if we're including them in Twig even if we're not directly referencing them as stories and switching from only loading the Twig file in page files to loading the entire template story to ensure all assets get loaded.

Can you look over these changes and see if anything jumps out as problematic? If not I'm going to roll out this approach to the other Gesso flavors as well.  Thanks!

### DIFF
--- a/source/00-config/storybook.global-data.yml
+++ b/source/00-config/storybook.global-data.yml
@@ -8,6 +8,7 @@ site_slogan: 'This is the site slogan.'
 section: 'Section'
 page_title: 'Page Title'
 title: 'Title'
+is_published: true
 show_admin_info: false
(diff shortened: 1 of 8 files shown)


dcmouyard (approved):
These changes look good to me.

PR #785, opened 2024-01-23
Macro to Normalize File Sizes

tomealter:
Converts bytes to KB, MB, GB etc.


dcmouyard (changes requested):
@tomealter Can you add an empty line at the end to match our coding standards? Otherwise this looks good to me.

Should we add an example of this macro getting used in the theme?

tomealter:
@dcmouyard Any recommendation as to where I should put an example? The file component seems obvious, but I wasn't sure if that would interfere with the default behavior of media document. Drupal handles the conversion automatically by default, it's just when we're accessing the raw value that we need to convert.

dcmouyard (approved):
I’m leaning towards not including an example. The closest example i could think of would be a file view mode for document media types, but not sure it’s worth including that out of the box.

PR #794, opened 2024-02-23
Sortable table

dcmouyard:
This PR adds a sortable table component, with the script ported over from USWDS.


kmonahan:
@mdrayer @ddhogan @mrclay Any of you want to review this one (specifically source/03-components/table/table--sortable.es6.ts)? I reviewed when Dan built this initially for client work so I think the TypeScript would benefit from a fresh set of eyes.

source/03-components/table/table--sortable.es6.ts
```diff
@@ -0,0 +1,295 @@
...
+const SORT_ALL_COLUMNS = true;
+
+/**
+ * Gets the data-sort-value attribute value, if provided — otherwise, gets
+ * the innerText or textContent of the child element (HTMLTableCellElement)
+ * at the specified index of the given table row.
+ *
+ * @param {number} index
+ * @param {HTMLTableRowElement} tr
+ * @return {string | number | null}
```
mdrayer: Is this best practice? With TS, this seems unnecessary as the types are already declared. If we still want to do this, the params are out of order: `tr` is second in the JSDoc but first in the function.

source/03-components/table/table--sortable.es6.ts
```diff
@@ -0,0 +1,295 @@
...
+ */
+const getCellValue = (
+  tr: HTMLTableRowElement,
+  index: number
+): string | number | null =>
+  (tr.children[index] as HTMLTableCellElement).getAttribute(SORT_OVERRIDE) ||
+  (tr.children[index] as HTMLTableCellElement).innerText ||
+  (tr.children[index] as HTMLTableCellElement).textContent;
```
mdrayer: Should we be casting like this instead of doing some sort of check? For instance, we could be doing something like the following for better type safety:
```ts
  const item = tr.children[index];
  if (item instanceof HTMLTableCellElement) {
    return (
      item.getAttribute(SORT_OVERRIDE) || item.innerText || item.textContent
    );
  }
  // Item not HTMLTableCellElement.
  return null;
```

source/03-components/table/table--sortable.es6.ts
```diff
@@ -0,0 +1,295 @@
...
+
+/**
+ * Compares the values of two row array items at the given index, then sorts
+ * by the given direction.
+ * @param {number} index
+ * @param {boolean} isAscending
+ * @return {number}
```
mdrayer: Recommend removing JSDoc types as they don't match up with the actual types.

source/03-components/table/table--sortable.es6.ts
```diff
@@ -0,0 +1,295 @@
...
+  updateSortLabel(header);
+
+  const table = header.closest(TABLE);
+  const headerRow = header.parentNode;
+
+  if (!table || !headerRow) {
+    return false;
+  }
+
+  const tbody = table.querySelector('tbody') as HTMLTableSectionElement;
```
mdrayer: I think this is unnecessary. Instead, we can simply do:
```ts
const tbody = table.querySelector('tbody');
```
Note that this makes the type of `tbody` possibly `null` (`HTMLTableSectionElement | null`), so we should be doing a check that it isn't `null` somewhere before using it:
```ts
if (!tbody) {
  return false;
}
```

source/03-components/table/table--sortable.es6.ts
```diff
@@ -0,0 +1,295 @@
...
+  const siblingHeaders = headerRow.children as HTMLCollection;
```
mdrayer: Casting is not needed here, as `siblingHeaders` will already be of type `HTMLCollection`.

source/03-components/table/table--sortable.es6.ts
```diff
@@ -0,0 +1,295 @@
...
+ * @param {HTMLTableCellElement} sortedHeader
+ */
+const updateLiveRegion = (
+  table: HTMLTableElement,
+  sortedHeader: HTMLTableCellElement
+): void => {
+  const caption = table.querySelector('caption')?.innerText;
+  const sortedAscending = sortedHeader.getAttribute(SORTED) === ASCENDING;
+  const headerLabel = sortedHeader.innerText;
+  const liveRegion = table.nextElementSibling as HTMLElement;
```
mdrayer: Suggest not casting and instead doing a check, as `liveRegion` could possibly not be found and be `null`:
```ts
const liveRegion = table.nextElementSibling;
if (!liveRegion || !(liveRegion instanceof HTMLElement)) {
  return;
}
```

source/03-components/table/table--sortable.es6.ts
```diff
@@ -0,0 +1,295 @@
...
+ * and vice-versa).
+ *
+ * @param {HTMLTableCellElement} header
+ * @param {boolean?} isAscending
+ */
+const toggleSort = (
+  header: HTMLTableCellElement,
+  isAscending: boolean
+): void => {
+  const table = header.closest(TABLE) as HTMLTableElement;
```
mdrayer: Suggest removing the casting and instead doing a check alongside `!table` down below:
```ts
const table = header.closest(TABLE);
...
if (!table || !(table instanceof HTMLTableElement)) {
  ...
}
```

source/03-components/table/table--sortable.es6.ts
```diff
@@ -0,0 +1,295 @@
...
+    tables.forEach(table => {
+      // Add sortable attribute to all column headers if wanted.
+      if (SORT_ALL_COLUMNS) {
+        table.querySelectorAll<HTMLTableCellElement>('thead th').forEach(th => {
+          th.setAttribute(SORTABLE_ATTRIBUTE, '');
+        });
+      }
+
+      // Add aria-live region for sort announcements.
+      const liveRegion = table.nextElementSibling as HTMLElement;
```
mdrayer: Removing casting as we do a check another on the next line to see if it is `null`.

source/03-components/table/table--sortable.es6.ts
```diff
@@ -0,0 +1,295 @@
...
+        table.querySelectorAll<HTMLTableCellElement>(SORTABLE_HEADER);
+      sortableHeaders.forEach(header => {
+        createHeaderButton(header, imagePath || '');
+      });
+
+      const sortButtons: NodeListOf<HTMLButtonElement> =
+        table.querySelectorAll(SORT_BUTTON);
+      sortButtons.forEach(button => {
+        button.addEventListener('click', event => {
+          const target = event.target as HTMLElement;
```
mdrayer: Remove casting and do a check:
```ts
const target = event.target;
if (!(target instanceof HTMLElement)) {
  return;
}
```

source/03-components/table/table--sortable.es6.ts
```diff
@@ -0,0 +1,295 @@
...
+          const targetHeader = target.closest(
+            SORTABLE_HEADER
+          ) as HTMLTableCellElement;
+          event.preventDefault();
+
+          toggleSort(
+            targetHeader,
+            targetHeader.getAttribute(SORTED) === ASCENDING
+          );
```
mdrayer: Remove casting and do a check:
```ts
const targetHeader = target.closest(SORTABLE_HEADER);
...
if (targetHeader instanceof HTMLTableCellElement) {
  toggleSort(
    targetHeader,
    targetHeader.getAttribute(SORTED) === ASCENDING,
  );
}
```

mdrayer (changes requested):
A question on the use of JSDocs and a couple suggestions around the typings/castings.

source/03-components/table/table--sortable.es6.ts
dcmouyard: @mdrayer 
You suggested change resulted in a `Use object destructuring` linting error.

Updating to the following broke the script, but didn’t result in any errors.

```js
const { target } = event;

if (!(target instanceof HTMLElement)) {
  return;
}
```

dcmouyard:
@mdrayer Thanks for your review! I think I fixed all of the issues you brought up, except for one.

source/03-components/table/table--sortable.es6.ts
mdrayer: Hmm, what broke with the script? Maybe `HTMLElement` isn't the right type here. Could you try `Element` instead?
```ts
const { target } = event;
if (!(target instanceof Element)) {
  return;
}
```

kmonahan: Or `HTMLButtonElement` since that's how `sortButtons` are typed?

mdrayer: @kmonahan I am getting a spread of different elements on `event.target` on this event depending where I click:
- `<svg class="c-icon c-table__sort-icon">`
- `<button tabindex="0" class="c-table__header-button">`
- `<use xlink:href="images/sprite.artifact.svg#sorted"></use>`

I'd guess that sticking with the `Element` check is the way to go here, unless we want to either adjust the event target or use a different property of `event` for the check here.

kmonahan: Or use `currentTarget` instead of `target`, since that should always be the element the click event is fired on versus the one that triggered it initially?

kmonahan: Whatever @dcmouyard thinks makes sense here -- I think `Element` is fine too

mdrayer: Ah yep, `currentTarget` works as well here and is consistently giving me the `<button>` element.

dcmouyard: Thanks for your help! Got it working with
```js
const target = event.currentTarget;
if (!(target instanceof HTMLButtonElement)) {
  return;
}
```

kmonahan (approved):
Looks good to me! Will let @mdrayer have final approval.

mdrayer (approved):

PR #796, opened 2024-02-23
Attributes

### LINKED ISSUE #795: Node attributes are ignored, and "role" attribute is redundant
This two issues. 

In Gesso's node templates, we go through the effort of combining our classes with the `attributes` variable passed to the theme template, but then we just ignore the attributes variable, as it's not passed to the included template. A couple of examples

[Full display for node](https://github.com/forumone/gesso/blob/5.x/templates/content/node--full.html.twig#L75)

[Teaser display](https://github.com/forumone/gesso/blob/5.x/templates/content/node--teaser.html.twig#L75)

I propose we pass the attributes variable to the component template, and begin using that in the article tag, like core does. Is there any reason not to do this? I believe the fact that we do not do this currently interferes with things like contextual links.

Second issue, which is really a follow up to the first: 

We are still using a `role="article"` attribute hard-coded on the `article` tag within the [component template](https://github.com/forumone/gesso/blob/5.x/source/03-components/teaser/teaser.twig#L13).

Drupal core actually [removed this](https://www.drupal.org/project/drupal/issues/3117230) from their templates in 10.1. Because the default "role" value for "article" tags is "article". For supporting documentation, see the the "Implicit ARIA role" in Mozilla's [HTML documenation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article). I propose we remove this, which would effectively happen anyway from the first proposal I made.

dcmouyard:
Fixes #795.

This PR explicitly adds the `attributes` variable to all relevant Twig templates and removes the hard-coded `role="article"` attribute in the Teaser and Image Teaser components.


templates/field/field--media--remote-video.html.twig
```diff
@@ -38,6 +38,7 @@
 #}
 {% for item in items %}
   {% include '@components/video/video.twig' with {
+    'attributes': attributes,
```
kmonahan: Since we're looping through the field items at this point, should the attributes be `item.attributes` instead? Same question applies to all templates where we're trying to pass attributes into a template to render a field item.

templates/form/fieldset.html.twig
```diff
@@ -36,5 +36,6 @@
 {% endif %}
 
 {% include '@components/fieldset/fieldset.twig' with {
-  modifier_classes: attributes.addClass(classes).removeClass(remove_classes).class|join(' ')|trim,
+  'attributes': attributes,
```
kmonahan: Since we're passing the attribute classes in as modifier classes, I think the attributes here should be `attributes|without('class')` so we're not passing in the attribute classes twice. 

Same note applies to all templates where we're explicitly passing in attribute classes (with or without modifications) as modifier_classes.

kmonahan (changes requested):
I think `attributes` is getting passed in more than it needs to here, but I'm not bothered by that, since explicit mappings can be easier for a developer to understand, even if it's a bit redundant. Plus it sets up the templates for easy modifications of the attributes as needed.

Couple things that I called out that do need to be fixed here though, and a question. My suspicion from reading the PR is that we do want `item.attributes` and not `attributes` for the field templates, but I could be wrong.

templates/form/fieldset.html.twig
dcmouyard: Good point. I refactored this a bit to make it explicit when we’re adding modifier classes, and making sure we don’t add duplicate classes or attributes.

kmonahan (approved):

PR #797, opened 2024-02-26
Update Field component to be compatible with the Fences module

### LINKED ISSUE #790: Field Configuration Discussion [Drupal]
I'd like to discuss the feasibility of making greater use of configurable formatters that can provide more options to site builders.  A module like [Fences](https://www.drupal.org/project/fences) allow site builders to define a field tag in the display configuration. But the current Gesso theme's [field template](https://github.com/forumone/gesso/blob/5.x/templates/field/field.html.twig) is incompatible with the field template required by that module. Example of [Fences field template](https://git.drupalcode.org/project/fences/-/blob/3.x/templates/field.html.twig?ref_type=heads).

Forum One's approach to site building in D7 perhaps leaned a bit too heavily in doing everything in configuration. But I feel like we've over-corrected a bit since 8, making heavy use of twig templates, and needing to create a new one if we need any level of customization, when there are options to just set things in the field's display configuration.   

Is there a way we can set up Gesso to be compatible with a module like Fences? Is there any interest in that from FEDs? 

Edit: I will note that the [Text Field Formatter](https://www.drupal.org/project/text_field_formatter) module seems to work in this use case, as it's not dependent on using it's own field template.

dcmouyard:
Fixes #790.

This PR:

- Updates Field component to be compatible with the Fences module, which includes:
  - Renaming variables to match what Fences outputs
  - Update field template to have more flexibility around item markup and classes
- Does not require Fences module
- Does not change default markup of fields in Gesso
- Remove List Field variant template, since it can now be achieved by sending appropriate data to Field template.
- Update `field--no-wrapper.html.twig` template to only add wrapper markup if logged in _and_ there are contextual links.


johnbburg (commented):
I just checked out this branch in a test site, and tested it with the fences module, and it seemed to work as expected with all the fences module options. I didn't test it with multi-value fields, but I wouldn't expect that to be an issue from looking at the code. so LGTM! I'll leave it to another Gesso dev to officially approve this.

johnbburg:
Update, Testing multi-value fields, it does insert a comma between values, which does mess things up if you are trying to say use fences to get a multi-value field to output as an `<ol>` or `<ul>`, so if you are going to do something like that, the suggestions might be to check out something like the field_delimiter module.

dcmouyard:
> it does insert a comma between values, which does mess things up if you are trying to say use fences to get a multi-value field to output as an `<ol>` or `<ul>`

The comma separated field items is how our default template is set up, but we could easily switch it so that it doesn’t.

That change would definitely be something that current projects would have to be mindful of when updating to match the latest Gesso. It also makes the default field output look weird if the Field Delimiter module isn’t installed out of the box.

It’s also very easy to set the default field template to not output commas on specific projects, which would be helpful on projects that use the Fences module.

I’m leaning towards not updating core Gesso to remove the commas, but I could be convinced otherwise. Thoughts?

johnbburg:
Sorry for the late reply. I'm inclined to not make assumptions on behalf of the user, but it looks like Corey whipped up an elegant solution.

PR #799, opened 2024-02-26
Add External Link component

dcmouyard:


kmonahan (changes requested):
Rather than hard-code the various allowed domains and hosts here and require a dev to comb through the code to change them each time, it would be nice if the various settings could be passed in. One option would be to set the external link functionality up as a class or separate module that can then have options passed in on creation. (The menus are one example of this). That still requires manual updating on the part of a developer, but at least it's contained in one place. Another option would be to manage this through the theme settings, similar to the button field formatter and the back to top. That's a little more work to set up, but would let us or the client add or remove domains in the future. Third option I can think of would be to declare the relevant variables as constants at the top of the file. Similar to the first, requires manual updating in code, but likewise makes it obvious what you need to change where.

My personal preference is for the second option (theme settings), because then you could mock up the settings in Storybook rather than putting example domains in the file. Since this is included as part of the global theme library, it'll automatically be added to sites unless we remove it, so it would be nice if we could start with sensible defaults (social media urls or no urls as allowed domains, current site as the only internal url) and then add to it as needed for projects that have additional requirements.

Love the idea of getting this in Gesso though! I've used your code on projects before, and this request comes up fairly frequently.

Marking as request changes, but it's definitely "request changes" not "demand changes" here.

dcmouyard:
I really like the theme settings approach, with an additional option to enable/disable external links. I don’t think that work blocks this PR, unless we want to fully bake this before merging.

if this PR isn’t merged before I get started on this, then I’ll just continue on this branch.

coreylafferty:
@dcmouyard Let's move some of this to theme settings before merging.

coreylafferty (changes requested):
Per KJ's suggestion, let's move some of this to theme settings before merging.

dcmouyard:
I’ve updated this PR to use theme settings to control external link icons.

includes/html.inc
```diff
@@ -32,15 +32,43 @@ function gesso_preprocess_html(array &$variables) {
...
 function gesso_page_attachments_alter(array &$attachments) {
   $include_back_to_top = theme_get_setting('include_back_to_top') ?? TRUE;
-  if ($include_back_to_top) {
-    $attachments['#attached']['drupalSettings']['gesso']['backToTopThreshold']
-      = theme_get_setting('threshold', 'gesso');
-    $attachments['#attached']['drupalSettings']['gesso']['backToTopSmoothScroll']
-      = theme_get_setting('smooth_scroll', 'gesso');
+  $add_external_link_icons = theme_get_setting('add_external_link_icons') ?? FALSE;
+
+  if ($include_back_to_top || $add_external_link_icons) {
```
kmonahan, line 37: I don't think you need the outer if statement here, since the two don't have any shared code.

source/03-components/external-link/external-link.es6.ts
```diff
@@ -0,0 +1,52 @@
...
+        if (
+          link.host === domain ||
+          link.host.endsWith(`.${domain}`)
+        ) {
+          external = false;
+        }
+      });
+
+      if (allowedLinks.includes(link.href.toLowerCase())) {
+        external = false;
```
kmonahan, line 32: Another way to do this would be to return as soon as something sets external to false, since if once we know the link is not external, there's no reason to do the other checks.

```
if (link.host === window.location.host) {
  return false;
}
// Other checks
return true;
```

source/03-components/external-link/external-link.es6.ts
```diff
@@ -0,0 +1,52 @@
...
+      }
+
+      return external;
+    }
+
+    externalLinks.forEach(link => {
+      if (link.hasAttribute('href') && linkIsExternal(link)) {
+        const accessibleLabel = Drupal.t(exitDisclaimer);
```
kmonahan: I believe the `Drupal.t()` needs to go around `exitDisclaimer` where it's defined, so you're translating a string, not a variable. I find the documentation confusing, but I think that's what https://www.drupal.org/docs/8/api/translation-api/overview#s-translation-in-javascript-files is saying

kmonahan (changes requested):
The only thing that needs to be addressed here is the use of `Drupal.t()` around dynamic content. The other comments are minor optimizations that can be addressed or not as you feel like.

includes/html.inc
dcmouyard: I believe they both share this:
```    /** @var \Drupal\Core\Render\Renderer $renderer */
    $renderer = \Drupal::service('renderer');
    $themeConfig = \Drupal::config('gesso.settings');
    $renderer->addCacheableDependency($attachments, $themeConfig);
```

source/03-components/external-link/external-link.es6.ts
dcmouyard: The problem I ran into using the `return: false` approach was with the following:
```
allowedDomains.forEach(domain => {
  if (
    link.host === domain ||
    link.host.endsWith(`.${domain}`)
  ) {
    external = false;
});
```
I need to loop though all domains and run the check. Won’t `return false;` just exit this arrow function instead of the parent function?

source/03-components/external-link/external-link.es6.ts
dcmouyard: Good catch! I pushed up this fix.

kmonahan (approved):

kmonahan:
@coreylafferty Marked this one as approved.

coreylafferty:
When reviewing, see if icon wrapping alone can be fixed or should be a new issue for icon/icon-link

coreylafferty:
@dcmouyard @kmonahan I saw some typing errors when compiling this locally. 
```
Property 'item' is missing in type 'Element[]' but required in type 'NodeListOf<HTMLAnchorElement>'.
```

Copilot suggested the changes I've made in this PR: https://github.com/forumone/gesso/pull/940

That seemed to fix the errors and the functionality seems to work in my testing. If that PR looks ok to you, I can go ahead and move forward with this and merge both PRs.

coreylafferty (approved):

PR #800, opened 2024-02-26
Add RTL support to list mixins

dcmouyard:


kmonahan:
@dcmouyard If we're going to update these mixins (which makes sense to me), I'd suggest switching them to CSS logical properties, so we don't need the RTL check. That's the approach we took in Gesso for React: https://github.com/forumone/nextjs-project/blob/1.x/source/00-config/mixins/list.css

kmonahan (changes requested):
I'm going to request changes, but if there's a reason you feel CSS logical properties aren't the right approach here, let's chat!

dcmouyard:
I like switching to logical properties, so I updated all of the list mixins to use that approach.

kmonahan (approved):

PR #801, opened 2024-02-26
Pager updates

### LINKED ISSUE #718: Consider removing role="presentation" on ... in pager
A couple of issues with our default markup came up on a recent project.

Corinne S. did some screenreader testing and found that while having the role does successfully remove the ellipsis from the accessibility tree and causes the screenreader to read the ellipsis as not part of the pager list, it also messes up the numbering.
> The link for “Page 3” is labeled 4 of 6, but when you advance to the ellipsis, it’s labeled 5 of 7. The final page is labeled 5 of 6, and the link for “next page” doesn’t get a list item number. 

Secondly, SiteImprove flags the role attribute on a list item as incorrect. Based on both Corinne S.'s and my research, we think this is a false report from SiteImprove (list items can have a role), but it's still worth noting because it takes additional dev and QA time to sort out false positives and communicate that to the client.

On the other hand, it looks like Olivero's default pager.html.twig template does still keep the role="presentation". So some additional testing and research might be warranted before we decide how to update.

dcmouyard:
Fixes #718.

This PR:

- Wrap markup in `{% apply spaceless %}`
- Remove unnecessary `role="navigation"` attribute on `<nav>`
- Remove unnecessary visually hidden and aria-hidden markup already covered by `aria-label` attribute
- Shorten accessible link labels (e.g. Go to last page -> Last page)
- On ellipsis list items, replace `role="presentation"` with `aria-label="ellipsis indicating non-visible pages"`
- Remove overridden focus styles in links to match the rest of the site

### DIFF
--- a/source/03-components/pager/pager--mini/pager--mini.twig
+++ b/source/03-components/pager/pager--mini/pager--mini.twig
@@ -27,48 +27,47 @@
 {% endset %}
 
 {% if pager_items.previous or pager_items.next %}
-  <nav {{ add_attributes({
-    'class': classes,
-    'aria-label': heading|default('Pagination'),
-    'role': 'navigation'
-  })}}>
-    <ul class="c-pager__items js-pager__items">
+  {% apply spaceless %}
+    <nav {{ add_attributes({
+      'class': classes,
+      'aria-label': heading|default('Pagination')
+    })}}>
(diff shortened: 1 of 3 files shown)


kmonahan (approved):
Awesome. Love that solution for the ellipses.

PR #812, opened 2024-03-08
[757] Prevent fonts from being emitted multiple times

### LINKED ISSUE #757: Update React Webpack with custom fonts
Currently, the webpack React config emits its own copy of any font files uploaded to the theme. Better would be for it to reference the existing font files output by the theme build.

kmonahan:
Closes #757 by switching the emitted fonts to being identified by name rather than hash (which also makes /dist/fonts easier to read).

Applies to GUSWDS.

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -154,7 +154,7 @@ module.exports = {
         exclude: ['/node_modules/'],
         type: 'asset/resource',
         generator: {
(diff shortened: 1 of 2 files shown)


dcmouyard (approved):
My previous error went away after I restarted Docker Desktop.

PR #817, opened 2024-03-20
Storybook 8 update

kmonahan:
Updates to Storybook 8. Note that this PR depends on and includes the work in https://github.com/forumone/gesso/pull/810 to ensure that Storybook 8 was set up to work with SWC.


kmonahan:
Also applies to GUS. Will open a separate PR on nextjs-project for Gesso fo React.

PR #825, opened 2024-03-21
refactored sidebar layout

coreylafferty:
The current markup/CSS for the sidebar layout appears to fail under some conditions. For example, the content area does not properly size if there is only a single right sidebar and if grid is not supported, the flexbox rules don't always behave as expected. I've refactored some of the logic and CSS to fix this.


kmonahan:
> if grid is not supported, the flexbox rules don't always behave as expected

No issue with refactoring the layout in general, but is grid not being supported a use case we still need to handle? It's [supported by all modern browsers](https://caniuse.com/?search=grid). I think we can drop the fallback here and just make sure that the grid display does all the things it should do.

PR #829, opened 2024-03-26
PR#797 tweak - removes separator from particular item wrapper elements for Fences

coreylafferty:
I've added an easily modifiable list of item wrapper elements to the default field template to disable field item separation if Fences is used to switch to something where commas would be disruption.

@dcmouyard @johnbburg What do you think of these changes?


dcmouyard (approved):
Looks good to me! 👍

PR #830, opened 2024-03-26
PR#808 tweaks - refactored refactored task menu a bit

coreylafferty:


coreylafferty:
@dcmouyard I noticed an issue where Pages/Page was not rendering (that's actually true in the current release of Gesso as well). I tracked it down to the use of the Article component in Storybook and these tabs/unpublished message.  

It never sat well with me that we had the tabs in the Article component AND in the homepage/landing page templates (because they don't use the Article component).  I refactored this a bit and moved the tabs/unpublished messaged into its own template, which other templates reference. This fixed the broken page and also seems more straightforward to me. Let me know if you agree.

PR #834, opened 2024-04-04
GitHub pages fix

kmonahan:


source/images/_sprite-source-files/sprite.cjs
```diff
@@ -0,0 +1,9 @@
+// Require all SVGs in the sprite so that they are processed through
+// webpack.
+function importAll(r) {
+  r.keys().forEach(key => {
+    console.log(key);
```
dcmouyard: Is this `console.log()` statement supposed to be here?

dcmouyard (commented):
The code changes look good to me (with the exception of a `console.log()` statement) but I’m not sure how to test this.

source/images/_sprite-source-files/sprite.cjs
kmonahan: Ah, good catch. Nope, I'll take that out. Otherwise, this can't be fully tested until it's deployed and GitHub Actions run, but if `npm run build-storybook` doesn't do anything weird and you're able to see the icon sprite in the generated `storybook/images` directory, then everything's working as it should.

kmonahan: `console.log()` removed!

coreylafferty (approved):

PR #842, opened 2024-04-18
[841] Allow stories to be imported for side effects

### LINKED ISSUE #841: JSX imports no longer include CSS & JS on page if not used
When updating from Gesso 5.2.8 to 5.3.2, I noticed that importing the stories file from another component no longer loads the CSS and JS for that component unless it’s rendered somewhere within the stories file. My hunch is that this has to do with updating to Storybook 8.

Using the homepage template as an example, in `source/04-templates/homepage/homepage.stories.jsx`:

```jsx
// Importing components to ensure their assets get loaded in Storybook when they
// get referenced since Drupal loads them as a library.
import { Default as HeroBGImage } from '../../03-components/hero-bg-image/hero-bg-image.stories.jsx';
```

The stylesheet for this component no longer gets loaded. Example: https://forumone.github.io/gesso/?path=/story/templates-homepage--homepage

Replacing `import { Default as HeroBGImage } from '../../03-components/hero-bg-image/hero-bg-image.stories.jsx';` with `import '../../03-components/hero-bg-image/hero-bg-image.scss';` fixes this specific issue, but requires us to manually update these dependencies everywhere they’re used rather than just in the original component stories file. It’s also more cumbersome when importing components that themselves import other components.

kmonahan:
Closes #841 . Explicitly marks `.stories.jsx` files as having side effects, so they won't be tree-shaken. That fixed the issue, but I went ahead and also updated the import method for stories that are imported just for their assets, since those should be using side effect imports anyway.


coreylafferty:
@kmonahan I'm getting a 'Drupal not defined' error after compiling all this. I don't think this is related, but do a pull on this branch b/c I updated it with the latest changes from 5.x-RC just in case there's something in there that caused this.

Untitled

kmonahan:
I'll take a look. Not sure at first glance why it would be related to the work here, but one never knows.

kmonahan:
Hm, still not totally sure why that error was appearing, but it seems like Storybook was no longer importing files with side effects that weren't called out in package.json, even though they were imported correctly. Maybe the assumption is that if `sideEffects` exists in package.json and a file is not listed in it, the file can be removed if doesn't appear to be used. (Interestingly, when I tested this originally, I was testing using `ddev gesso build`, since I thought tree-shaking only occurred in production build. But that seemed to be working yesterday, while when I fired up `ddev gesso watch`, I saw the error you were getting. So I'm still a bit puzzled.)

I tried a different approach and added sideEffects to the rule in the webpack config rather than to package.json, and that seems to be working better.

coreylafferty (approved):

PR #847, opened 2024-04-29
[ISSUE-793] Add hero paragraph template with inline media

### LINKED ISSUE #793: Adding template for inline hero to implement with Drupal Project
Implementing the hero component on Drupal Project and I'm wondering if a template would be welcomed here.

Here is the related [issue](https://github.com/forumone/drupal-project/issues/97) in Drupal Project and the corresponding [PR](https://github.com/forumone/drupal-project/pull/103).

Here is a [demo](https://f1-dev-f1-dp-demo.pantheonsite.io/example-page). Scroll to the bottom.

Beyond providing a template I also update a rule for the inline hero to include the use of `figure` since that is what is used out of the box with a media entity/image file type. Both the template and the tweak to the style can be viewed in this [commit](https://github.com/forumone/drupal-project/commit/713d972a06856af0e99449748ec2ed36a8ca919c). I am happy to create PR here if this change is welcomed.

gbeezus:
Resolves #793

### DIFF
--- a/source/03-components/hero-bg-image/hero-bg-image.scss
+++ b/source/03-components/hero-bg-image/hero-bg-image.scss
@@ -11,6 +11,8 @@ $hero-bg-image-bp: gesso-breakpoint(desktop) !default;
   position: relative;
   text-align: center;
 
+  > figure,
+  > figure img,
   > img {
     height: 100%;
     left: 0;
@@ -18,6 +20,24 @@ $hero-bg-image-bp: gesso-breakpoint(desktop) !default;
     position: absolute;
     top: 0;
     width: 100%;
+  }
+
+  &.has-content-left {
+    text-align: left;
+
+    @include breakpoint($hero-bg-image-bp) {
+      padding-left: 10%;
+      padding-right: 35%;
+    }
+  }
+
+  &.has-content-right {
+    text-align: left;
+
+    @include breakpoint($hero-bg-image-bp) {
+      padding-left: 35%;
+      padding-right: 10%;
+    }
   }
 
   &.has-overlay {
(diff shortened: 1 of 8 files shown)


dcmouyard:
@gbeezus What’s the best way to get the new paragraph in https://github.com/forumone/drupal-project/pull/103 into my local environment?

gbeezus:
Not sure I completely understand the question @dcmouyard . I imagine you want to test the changes as cleanly as possible. If you have your own version of drupal-project that you are running tests off of you could checkout out the branch with the changes [ISSUE-97-hero-paragraph](https://github.com/forumone/drupal-project/tree/ISSUE-97-hero-paragraph) and then copy the module to your current local install.

Alternatively, we are keeping a separate demo branch (mind it does not have the latest version of gesso) but that may also be a good starting point. [drupal-project-demo](https://github.com/forumone/drupal-project-demo). What I've been doing is: 
- branch off of main `git checkout -b [branch-name]`
- `git fetch upstream`
- `git merge --no-ff upstream/ISSUE-97-hero-paragraph`
- Then start the project
- enable the module

Maybe that's all a bit clunky but we can find ways to refine testing around all of this.

Let me know if there is anything I can clarify

dcmouyard (approved):
@gbeezus Thanks for the tips on getting that module locally!

Everything worked, but I made some updates to match our coding standards.

@coreylafferty @kmonahan @tomealter This PR can be merged as-is, but I’d really like to update our two hero components so that we can easily switch between them in this paragraph template. Any objection with me doing that work in this PR?

kmonahan:
@dcmouyard Definitely no objection to you doing that work. I think you should branch off this work though and open a separate PR with your changes (either onto this branch or onto 5.x-RC, depending on when this gets merged). Seems like that would be a little easier to review, since I don't think your planned work needs to block this PR being added to Gesso (unless you feel like it should?)

PR #852, opened 2024-05-02
includes media layout in image teaser

coreylafferty:
Instead of duplicating the media layout markup in the image-teaser twig file, this includes the layout.


dcmouyard (approved):
LGTM 👍

PR #855, opened 2024-05-07
[683] Adds modal component to gesso

### LINKED ISSUE #683: Create a modal component for Gesso

tomealter:
@coreylafferty @dcmouyard @kmonahan First stab at the modal component.

It uses `<dialog>` as the base element and the corresponding `showModal()` and `close()` methods for operation.

There is currently an optional variable for "Open by default" which automatically opens the modal on page load.

Let me know if there are other options we should consider.

I was assuming that the modal size and overlay color/effect/animation would be custom on a project by project basis. So I just included some simple styling with comments to point out where these can be updated.


source/03-components/modal/modal.es6.js
```diff
@@ -0,0 +1,112 @@
+import Drupal from 'drupal';
+import once from 'once';
+import KEYCODE from '../../00-config/_KEYCODE.es6';
```
dcmouyard: Do we still need this? Can’t we rely on `key`?

source/03-components/modal/modal.yml
```diff
@@ -0,0 +1,8 @@
+---
+default_open: true
+id: '1234'
+title: Modal Title
+content: |-
+  <p>Irure cillum laborum proident in nulla duis sint cillum esse minim. Aliqua non ea reprehenderit aliqua labore dolor adipisicing et aliquip ea occaecat elit nostrud. Esse aliquip tempor ex tempor deserunt. Commodo sunt minim fugiat non.</p>
+  <p>Irure aliqua reprehenderit ut pariatur culpa consequat ad reprehenderit consequat exercitation ut commodo. Tempor et consectetur mollit aliquip Lorem duis voluptate cillum aliquip adipisicing do reprehenderit. Cupidatat dolor enim labore est. Aliquip dolor aliquip dolore pariatur consectetur cillum officia sunt anim et non.</p>
+  <a href="#0">Read More</a>
```
dcmouyard: Minor nitpick: this file should end with a newline.

Also, we should add this as a library in `gesso.libraries.yml`.

dcmouyard (changes requested):
@tomealter Thanks for putting this together! It worked perfectly for me. I just have a question about using `keyCode` and a minor coding standard nitpick.

source/03-components/modal/modal.twig
```diff
@@ -0,0 +1,23 @@
...
+
+<dialog
+  id="modal-{{ id }}"
+  {{ add_attributes({ 'class': classes })}}
+  {{ default_open ? 'data-default-open' : '' }}
+  aria-labelledby="modal-title-{{ id }}"
+  aria-describedby="modal-content-{{ id }}"
+>
+  <div class="c-modal__inner js-modal-inner">
+    <button class="c-modal__close-button js-modal-close" aria-label="Close modal" aria-controls="modal-{{ id }}" autofocus>Close Modal</button>
```
dcmouyard: We don’t need button text since we’re using `aria-label` and the text  should be translatable:

`{{ 'Close modal'|t }}`

dcmouyard (changes requested):

source/03-components/modal/modal.es6.js
```diff
@@ -0,0 +1,112 @@
...
+
+Drupal.behaviors.accordion = {
```
dcmouyard: This should be renamed to `modal`.

source/03-components/modal/modal.es6.js
```diff
@@ -0,0 +1,112 @@
...
+    );
+
+    // Function to show modal
+    const openModal = modal => {
+      modal.showModal();
+      // eslint-disable-next-line @typescript-eslint/no-use-before-define
+      modal.addEventListener('click', handleOverlayClick);
+      // Turn off scrolling on the body
+      document.body.style.overflow = 'hidden';
+    };
```
dcmouyard, line 39: We should add an event here that other scripts can attach listeners to. Something like:

```js
const event = new CustomEvent('openmodal');
modal.dispatchEvent(event);
```

source/03-components/modal/modal.es6.js
```diff
@@ -0,0 +1,112 @@
...
+      modal.removeEventListener('click', handleOverlayClick);
+      // Turn on scrolling on the body
+      document.body.style.overflow = '';
+    };
+
+    // Close modal on overlay click
+    const handleOverlayClick = event => {
+      if (event.target.closest(`.${MODAL_INNER_CLASS}`)) return;
+      closeModal(event.currentTarget);
+    };
```
dcmouyard, line 50: We should add an event here that other scripts can attach listeners to. Something like:

```js
const event = new CustomEvent('closemodal');
modal.dispatchEvent(event);
```

dcmouyard (changes requested):
Sorry, I keep finding more stuff as I use this on a project!

Question: should we convert this to TypeScript?

kmonahan:
Nah, I don't think this needs to be converted to TypeScript. Currently TypeScript is _supported_ in Gesso for Drupal but not _required_. If we want to convert to TypeScript across the board, that's probably a separate issue/discussion.

source/03-components/modal/modal.es6.js
kmonahan: New convention proposal: we prefix our behavior names with gesso, e.g. `Drupal.behaviors.gessoModal`. It occurs to me that `accordion` and `modal` are common enough terms that it's possible some other module could also define `Drupal.behaviors.modal`.

source/03-components/modal/modal.es6.js
```diff
@@ -0,0 +1,112 @@
...
+      'close-modal-button',
+      `.${MODAL_CLOSE_BUTTON_CLASS}`,
+      context
+    );
+
+    // Function to show modal
+    const openModal = modal => {
+      modal.showModal();
+      // eslint-disable-next-line @typescript-eslint/no-use-before-define
+      modal.addEventListener('click', handleOverlayClick);
```
kmonahan, line 33: Rather than disable the linter here, can you move `handleOverlayClick` above `openModal`?

source/03-components/modal/modal.es6.js
```diff
@@ -0,0 +1,112 @@
...
+      // Turn off scrolling on the body
+      document.body.style.overflow = 'hidden';
+    };
+
+    // Function to close modal
+    const closeModal = modal => {
+      modal.close();
+      // eslint-disable-next-line @typescript-eslint/no-use-before-define
+      modal.removeEventListener('click', handleOverlayClick);
```
kmonahan, line 44: Ideally, you wouldn't disable the linter here either, but this one might be inevitable, since handleOverlayClick classes closeModal in return.

source/03-components/modal/modal.es6.js
```diff
@@ -0,0 +1,112 @@
...
+          else if (document.activeElement === lastFocusableElement) {
+            // Add focus for the first focusable element
+            firstFocusableElement.focus();
+            e.preventDefault();
+          }
+        } else if (e.key === 'Escape' || e.keyCode === KEYCODE.ESCAPE) {
+          // Close modal on escape key press
+          e.preventDefault();
+          // eslint-disable-next-line @typescript-eslint/no-use-before-define
+          closeModal(element);
```
kmonahan, line 80: Why did the linter need to be disabled here? It looks to me like closeModal() has already been defined?

source/03-components/modal/modal.es6.js
tomealter: Yeah, I like that idea @kmonahan.
Also, nice catch @dcmouyard , I have a bad habit of copy/pasting the first few lines of another JS and forgetting to rename them 🤦

source/03-components/modal/modal.es6.js
tomealter: Yeah, having one disable seems inevitable.. I did move `handleOverlayClick` above the open and close functions though, which means we only need one disable instead of two.

source/03-components/modal/modal.twig
tomealter: Button text removed.

source/03-components/modal/modal.yml
tomealter: Yml file fixed and library entry added.

source/03-components/modal/modal.es6.js
tomealter: Removed references to _KEYCODE.es6

source/03-components/modal/modal.es6.js
tomealter: Fixed.

source/03-components/modal/modal.es6.js
tomealter: @dcmouyard Mind showing me what you're thinking here? Feel free to add to this branch.

source/03-components/modal/modal.es6.js
dcmouyard: Me too! 😄

source/03-components/modal/modal.es6.js
dcmouyard: @tomealter I just added a commit with these new custom events.

tomealter: @dcmouyard This looks great!

tomealter:
@dcmouyard @kmonahan @coreylafferty called out items have been fixed.

kmonahan (approved):

dcmouyard (approved):
The changes look good to me.

Using this on a project, I ran into an issue where I needed to open the modal in another script. Using `showModal()` worked, but it didn’t handle everything that the new custom `openModal()` does, and I couldn’t call `openModal()` from the other script.

I wonder if we should refactor this to create a new ModalDialog Class that extends HTMLDialogElement? Not sure if this work should be done in this PR or a new one.

kmonahan:
@dcmouyard Definitely a different PR. I'm also not sure extending HTMLDialogElement is the best way to do it (but open to being convinced otherwise!) I know you just did it somewhere and assume there was good reason that warranted it, but in terms of what we want to put as standard in Gesso, I'd want to first explore whether we can find something that doesn't require overriding HTML elements. A couple initial thoughts: a custom event the modal listens for that can be fired to open the modal or wrapping the modal code in a class so the actual modal element is a property of the modal. For now, I think it's worth opening an issue for but I think it can be future work.

PR #856, opened 2024-05-08
849: Accessible accordion attributes

### LINKED ISSUE #849: Accordion ARIA attributes
The accordion script should be updated to not add the `aria-expanded` attribute to the div that is expanding/collapsing.

dcmouyard:
Fixes #849. This PR:

- Removes the unnecessary attributes on accordion sections
- Updates the helper slide functions to account for `prefers-reduced-motion`


source/03-components/accordion/accordion-item.twig
```diff
@@ -27,9 +27,6 @@
     <div
       class="c-accordion-item__drawer js-accordion-drawer"
       id="{{ section_id }}"
-      role="region"
-      aria-labelledby="{{ button_id }}"
-      aria-expanded="true"
```
tomealter: @dcmouyard I was planning on just swapping out `aria-expanded` for `aria-hidden` and leaving the `role` and `aria-labeledby`. Is there no reason to keep any of it? For example: https://www.w3.org/WAI/ARIA/apg/patterns/accordion/examples/accordion/

dcmouyard: @tomealter I was following the example of the [USWDS Accordion](https://designsystem.digital.gov/components/accordion/) component. I know `aria-expanded` is problematic here, but I’m not sure whether `role` and `aria-labelledby` are as well.

I think `aria-hidden` would work here as long as we’re still hiding the content via `display: none` somewhere. Out of the box we are with `slideToggle()`, but if folks remove that on a project then they would need to address this.

tomealter:
@dcmouyard I pushed up some updates to the accordion JS to trim some fat and make things a little more consistent.

kmonahan (commented):
Haven't tested, but read through the code and changes here all look good to me.

dcmouyard:
@tomealter Your changes look good to me. I just pushed up a tweak, though, that:

- Renames `step_list` variable to `is_step_list` so it’s more clear that it’s a boolean.
- Only output the step list markup if it’s a step list.

kmonahan (approved):

PR #866, opened 2024-06-03
2024-06 Package Updates

kmonahan:
Updates everything to the next minor and/or patch versions and updates css-loader and style-loader to the new major versions. 

Most relevant/interesting changes:
- Latest version of stylelint-scss lints for anywhere that Sass global functions are used without modules, so fixed existing instances of those.
- We're now on Storybook 8.1


dcmouyard (approved):
Build process completed successfuly and everything looks like it’s working correctly.

Issue #838, opened 2024-04-18
Create call to action Storybook component and Drupal paragraph template


carsoncho:
A common component we have on all of our sites is a Call to Action component. The thought here would be that component would have at least the following fields:

- Heading
- Body text/copy
- Link
- (Optional) image?

I'm unsure if we need to provide designs or anything for that, but we feel this would be a useful component to include as part of our starter projects and thus part of Gesso.

coreylafferty:
@carsoncho We discussed this during the Gesso development meeting and the general consensus was that callouts vary wildly in design and are typically fairly easy to develop, so a starting one may not be that useful as it'd often be easier to start from scratch rather than undo and modify an existing one.  

That said, if designers ARE moving towards a more standardized approach to callouts or if their starting design templates have one (which might lead designers to stick somewhat closely to its design and layout), then it might be valuable to have an out of the box option.  

If one of those things is true and you can provide a design then we could discuss further. Thanks!

coreylafferty:
@carsoncho We didn't realize this was part of the Drupal starter paragraphs project when we discussed this earlier. In that case this makes more sense coming from that direction... we'd just need to give it very minimal design on the theming side, but I can do that as well as create the supporting templates once the fields are created on the backend. If you can point me to a PR or something for the backend portion, I can add supporting templates and styles to Gesso and help with a default template in the module if needed.

coreylafferty:
@carsoncho Did this paragraph type ever get created as part of drupal starter?

carsoncho:
@coreylafferty Yes, this component now exists as part of drupal-starter. This is included in the latest drupal-starter now ([https://github.com/forumone/drupal-project/tree/main/web/modules/starterkit/f1_call_to_action_paragraph](https://github.com/forumone/drupal-project/tree/main/web/modules/starterkit/f1_call_to_action_paragraph)). This follows the rough content model proposed above.

coreylafferty:
Next step -- create a Storybook "call to action" component with a very simple design and a Drupal paragraph template that matches the configuration (field machine names, etc.) in the Drupal Project (https://github.com/forumone/drupal-project/tree/main/web/modules/starterkit/f1_call_to_action_paragraph).  This is similar to what we've done for the Hero and Accordion paragraphs.

carsoncho:
Apologies, I didn't keep up with this issue's progress! Happy to see this is now implemented now. We can call this issue closed.

PR #886, opened 2024-07-17
Issue 838: Add CTA component

jbwagner:
Completes issue #838 

Adding a basic CTA component that matches with the Drupal project CTA paragraph.


source/03-components/_index.scss
```diff
@@ -34,3 +34,4 @@ $wysiwyg: false !default;
 @use 'view/view';
 @use 'view/views-view/views-view';
 @use 'view/views-view-unformatted/views-view-unformatted';
+@use 'call-to-action/call-to-action';
```
coreylafferty: @jbwagner Let's remove this and make the CSS file not a partial.  Then add it as a Drupal library.

source/03-components/call-to-action/_call-to-action.scss
```diff
@@ -0,0 +1,24 @@
+// @file
+// Component: c-call-to-action
+
+@use '00-config' as *;
+
+.c-call-to-action {
+  .l-grid {
+    align-items: center;
+  }
```
coreylafferty: @jbwagner Not a huge fan of adding non-call-to-action classes in here. Is there another way to achieve this? Maybe with an extra wrapper around the grid?

source/03-components/call-to-action/_call-to-action.scss
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.c-call-to-action__media {
+  @include breakpoint(gesso-breakpoint('tablet')) {
+    grid-column-start: 2;
+    order: 2;
+  }
+}
+
+.c-call-to-action__copy {
```
coreylafferty: @jbwagner Maybe call this 'text' instead of 'copy', just for consistency with other components?

source/03-components/call-to-action/call-to-action.twig
```diff
@@ -0,0 +1,28 @@
...
+  <div class="c-call-to-action__copy">
+    {% if heading %}
+      <h2>{{ heading }}</h2>
+    {% endif %}
+    {{ body }}
+    {% if link.url and link.text %}
+      <div>
+        <a href="{{ link.url }}" class="c-button">{{ link.text }}</a>
+      </div>
+    {% endif %}
```
coreylafferty: @jbwagner in your yaml, these are non-nested variables called 'title' and 'url'

source/03-components/mobile-menu/modules/_MobileMenu.es6.js
```diff
@@ -78,16 +78,20 @@ class MobileMenu extends OverlayMenu {
...
-    const childrenWithFor = [...blockClone.querySelectorAll('[for]')];
-    childrenWithFor.forEach((e) => {
+    const childrenWithFor = blockClone.querySelectorAll('[for]');
+    if (childrenWithFor.length) {
+    childrenWithFor.forEach(e => {
       const thisFor = e.getAttribute('for');
       e.setAttribute('for', `${thisFor}-mobile`);
-    });
+      });
+    }
```
coreylafferty, line 94: @jbwagner Where did these changes come from?

templates/paragraph/paragraph--call-to-action.html.twig
```diff
@@ -0,0 +1,7 @@
+{% include '@components/call-to-action/call-to-action.twig' with {
+  'media': content.field_cta_media|field_value is not empty ? content.field_cta_media|field_value : false,
+  'heading': content.field_cta_heading|field_value is not empty ? content.field_cta_heading|field_value : false,
+  'body': content.field_cta_body|field_value is not empty ? content.field_cta_body|field_value : false,
+  'url': content.field_cta_link|field_value[0]['#url'] is not empty ? content.field_cta_link|field_value[0]['#url'] : false,
+  'text': content.field_cta_link|field_value[0]['#title'] is not empty ? content.field_cta_link|field_value[0]['#title'] : false,
```
coreylafferty: @kmonahan Would you mind weighing in on the best way to call url/title from a link field?  Also, do we need add any cache catching to this paragraph template?  If you have time, it might be worth you taking a look at the existing paragraph templates (hero, accordion, etc.) in Gesso 5.x and seeing if any adjustments are needed there for cache catching or field value calls.

coreylafferty (changes requested):

source/03-components/_index.scss
jbwagner: Fixed [6dd6fd9](https://github.com/forumone/gesso/pull/886/commits/6dd6fd9ac15e196e3d10c525df7a77e293b0471c)

source/03-components/call-to-action/_call-to-action.scss
jbwagner: Added a modifier class to the grid component, should be a bit more self-contained [a8cd193](https://github.com/forumone/gesso/pull/886/commits/a8cd193de80ee2f4feea7c09dc6c934c7bb7fffe)

source/03-components/call-to-action/_call-to-action.scss
jbwagner: Changed. Naming things is hard. [437f903](https://github.com/forumone/gesso/pull/886/commits/437f9034add8a9ca842686fb7f10b48fe5548223)

source/03-components/call-to-action/call-to-action.twig
jbwagner: That was intended to be non-nested, something I changed when testing inside a Drupal install but seem to have missed when making those changes on this repo. [a16c5d9](https://github.com/forumone/gesso/pull/886/commits/a16c5d9c8f28b562529738131b4f0150d33fc758)

templates/paragraph/paragraph--call-to-action.html.twig
jbwagner: Ok based on what's in the Hero paragraph template, I swapped it to look in the `paragraph` object instead of the `content` render array. And that template doesn't have any caching checks/additions. But caching is never a bad thing to double-check, so yeah, is pulling from the paragraph ok? I feel like that's precisely when we'd need to add things to the cache list.

[7314900](https://github.com/forumone/gesso/pull/886/commits/731490071d152466f1c4f75784265772c2f47322)

kmonahan: If we're pulling the URL and text values out of the field, then we need to ensure that cache tags and any attributes on the link itself are not lost. (I think the content array is also where we'd want to pull from, so that may need to get fixed in the hero paragraph too. 🤔 ) Agreed that this probably applies to other paragraphs as well. I'll take a look -- will be after various headless things though.

kmonahan:
Opened [838-catch-the-cache](https://github.com/forumone/gesso/pull/890) with my changes. @jbwagner for some weird reason GitHub wouldn't let me tag you as a reviewer on the PR, but feel free to take a look.

coreylafferty (approved):

PR #887, opened 2024-07-19
Add medium constrain class

mrclay:


coreylafferty:
@mrclay the default l-constrain with no modifiers already uses the medium size, so this would be redundant.  Ok to close?

PR #889, opened 2024-08-05
2024-08 package updates

kmonahan:
A little more dramatic than usual as we've got some major/breaking changes this round:
## Sass Nesting
Most importantly, [Sass has changed how it handles nesting so that it is aligned with how CSS handles nesting](https://sass-lang.com/documentation/breaking-changes/mixed-decls/). That means that putting declarations after nested rules will fill up your terminal with deprecation warnings. You either need to move all of the declarations before the nested rules or opt in to the CSS nesting approach by wrapping your declarations following the nesting rule with `& {}`. (The link has some examples.) 

For us, this comes up whenever we've got a mixin that includes nested styling (either child selectors, ::before or ::after, or :hover, :focus, etc.). Since we can't then put all of the declarations before the nested rules (because some are in the mixin and some are in the selector that's using the mixin), we'll need to wrap subsequent styles in `& {}`. This does make them compile slightly differently, as previously, Sass would just move the declarations for us.

### Before
```
.fancy-link {
  @include link(#ff0, #0f0);
  font-size: 2rem;
  text-decoration: dotted;
}
```
compiled to
```
.fancy-link {
  color: #ff0;
  font-size: 2rem;
  text-decoration: dotted;
}

.fancy-link:hover {
  color: #0f0;
}
etc.
```

### After
```
.fancy-link {
  @include link(#ff0, #0f0);

  & {
    font-size: 2rem;
    text-decoration: dotted;
  }
}
```
and now compiles to
```
.fancy-link {
  color: #ff0;
}

.fancy-link:hover {
  color: #0f0;
}
etc.

.fancy-link {
  font-size: 2rem;
  text-decoration: dotted;
}
```

## Drupal Once
The package.json for `@drupal/once` has an error that seems pretty minor (missing the `./` in the exports targets) but Storybook 8.2 errors out now when hitting it. I've opened [an issue and MR on DO](https://www.drupal.org/project/once/issues/3466026) to hopefully get this fixed but in the mean time, since once.js is a single file and we're only using it in Storybook to stub `drupal/once` from core, I copied the script into our stubs folder.

## Other Noteworthy Changes
- sass-loader now uses the modern Sass API, which changes up the options. Most relevant for our purposes: `includePaths` is now `loadPaths`. [Full docs for new options](https://sass-lang.com/documentation/js-api/interfaces/options/)
- stylelint-config-sass-guidelines [adds back rules that had been removed in a previous version](https://github.com/bjankord/stylelint-config-sass-guidelines?tab=readme-ov-file#stylistic). If you update an existing project, you may need to fix some new linting issues and/or update your .stylelintrc.yml to turn those rules off.
- At least one dependency (glob) now requires Node 20


kmonahan:
Converting to draft until issues @coreylafferty and I found this morning are fixed:

- The copy of once.js throws type errors (not clear to me yet whether this is something that needs to be fixed in stubs/once.js or by fixing the .ts files where the errors are occuring but either way, that needs to not happen)
- Do some more investigation into whether we can disable the specific SCSS deprecation warning so existing projects won't need to update their SCSS and can just wait it out

coreylafferty:
Once.js issues have been resolved and we've decided to not use the opt-in nesting code for now, instead hiding the warnings for just this issue. Once Sass puts out a release using the new code ordering we'll make any necessary updates to our code (if there are any).

coreylafferty (approved):

PR #907, opened 2024-10-11
adds styles for extlink icon

coreylafferty:
This augments the default styling of the external link icon from the extlink module, making it match the color/size of its parent and generally look better next to lines of text.


kmonahan:
We've got https://github.com/forumone/gesso/pull/799 already underway, and I'd lean toward that over using the extlink module as our standard approach. So not sure we want to add extlink module-specific CSS here.

PR #916, opened 2024-11-14
adjusted content area of default page template

coreylafferty:
* Added a wrapper div around article content.
* While I think it makes sense for the default article content to be wrapped in the wysiwyg, our default page template uses article and generally a page will have its own wysiwyg or other paragraph components.  Due to that, I've used the content block to just pass content in directly without the wysiwyg wrapper.

### DIFF
--- a/source/03-components/article/article.twig
+++ b/source/03-components/article/article.twig
@@ -45,9 +45,11 @@
     </footer>
   {% endif %}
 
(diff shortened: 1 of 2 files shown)


kmonahan (approved):

kmonahan:
@kmonahan

PR #929, opened 2025-02-20
fixes empty header tag being created if view has no title

coreylafferty:
Currently if a view title is left empty, the H2 tag is still created for it (with no text).  This fixes that.


kmonahan (approved):

kmonahan:
@kmonahan

Issue #819, opened 2024-03-20
Fix mobile menu subnav toggle when using mega menu.


tomealter:
When mega menu is used as the primary menu type, the subnav toggles do not work initially.

tomealter:
PR created - https://github.com/forumone/gesso/pull/933

PR #933, opened 2025-03-13
819 - Fixes mobile menu subnav toggles when mega menu is used.

tomealter:
…nav toggles

### DIFF
--- a/source/03-components/mobile-menu/modules/_MobileMenu.es6.js
+++ b/source/03-components/mobile-menu/modules/_MobileMenu.es6.js
@@ -258,7 +258,7 @@ class MobileMenu extends OverlayMenu {
           `.${this.options.classPrefix}__subnav-toggle`
         );
         if (this.options.toggleSubnav) {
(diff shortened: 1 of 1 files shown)


tomealter:
Let me know if there are any concerns with this change. The problem is that when mega-menu is selected for main menu, the `nextElementSibling` is not a `ul`, it's a `div`.
Related to issue - https://github.com/forumone/gesso/issues/819

coreylafferty:
Also look and see if dropdown carets not showing is intentional.

kmonahan:
Took a quick look at this and I think the missing chevrons _is_ intentional since there are also not chevrons on the desktop version. The entire link is a button.

It's worth considering whether we want to add chevrons (I don't know why I didn't put them there...) but that's likely a separate bug/issue.

coreylafferty (approved):

coreylafferty:
Merging this fix since it is unrelated to the chevron discussion, but I did create an issue for that possible improvement.

PR #934, opened 2025-03-17
2025-03 package updates

kmonahan:
Minor and package updates, plus updates to @swc/cli, @storybook/addon-webpack5-compiler-swc, webpack-cli, react-refresh, react, and react-dom.

Most noteworthy update is to React and React DOM. The previously deprecated global JSX namespace has now been removed, so you definitely have to import it if you're using React with TypeScript, and the syntax for rendering a React component on a page has changed slightly. Even still, changes are pretty minor here.


tomealter:
@kmonahan Gave this a look over, the theme builds fine initially. Since you called out the React changes, I tested the build there too and I did notice an issue.
If you make any changes to the `source/07-react/Example/Example.tsx`, the watch build throws the following error:
`ERROR in TS1: TSC compiler crashed: Debug Failure. False expression. Error: Debug Failure. False expression.`

coreylafferty (changes requested):
Tommy requests changes.

kmonahan:
@tomealter Looks like the issue was related to https://github.com/microsoft/TypeScript/issues/60692 and the same workaround worked for me locally. Please retest and see if error is fixed on your end as well!

coreylafferty:
@kmonahan I went to test this and install failed saying the lock file was not in sync with the package file. Not sure if that was overlooked or if something new has been recently released -- but I created this PR for the lock changes. Can you confirm these look ok? If so, I'll more forward with merging both PRs.  

https://github.com/forumone/gesso/pull/936

coreylafferty (approved):

PR #940, opened 2025-03-28
fixes for typing error

coreylafferty:
I saw some typing errors when compiling this locally.

```
Property 'item' is missing in type 'Element[]' but required in type 'NodeListOf<HTMLAnchorElement>'.
```

Copilot suggested the changes I've made in this PR.

That seemed to fix the errors and the functionality seems to work in my testing. If that PR looks ok to you, I can go ahead and move forward with this and merge both PRs.


kmonahan:
I put my approach in https://github.com/forumone/gesso/pull/941. Since once only returns Element[], does a quick filter to removing anything that isn't an HTMLAnchorElement (which there shouldn't be, but TypeScript would like to argue that we don't _know_ that because someone could change the selector string), so that going forward, externalLinks will safely only include HTMLAnchorElements.

Issue #945, opened 2025-04-04
Fix or remove Gesso Drush commands


dcmouyard:
Current implementation of Drush commands to copy & rename the theme is broken. We should either fix it or remove this functionality.

dcmouyard:
Duplicate of #872.

kmonahan:
Reopening based on discussion in https://forumone.slack.com/archives/CPPHLB318/p1760391805747749 as well as seeing a couple other issues. It seems like the drush command functionality is causing issues on projects that use Phpstan. Given its low use (it's been broken for...a year?), I think we should remove it.

kmonahan:
@coreylafferty , @dcmouyard , what do you think?

PR #946, opened 2025-04-04
806: Fix phpcs and phpstan errors

### LINKED ISSUE #806: PHPStan issue
When setting up a new theme there are some common PHPStan things that come up and should be fixed.

Screenshot 2024-03-06 at 9 30 23 AM

dcmouyard:
Fixes #806.

This PR fixes all phpcs and phpstan errors except for the following, which can be added to `phpstan-baseline.neon` and ignored on projects:

- Function has parameter with no value type specified in iterable type array.
- Function return type has no value type specified in iterable type array.
- Method has parameter with no value type specified in iterable type array.
- Method return type has no value type specified in iterable type array.
- Call to method on an unknown class.
- Call to an undefined method.

The last two are problems with Gesso Drush commands, which will be tackled in #945.


kmonahan (approved):
Looks good to me! Wonder if we should discuss these changes at a future FED meeting or something as well so everyone's aware of what our current standards are, especially around typing.

coreylafferty:
@dcmouyard @kmonahan do we need to create a related PR to drupal-project to add those exceptions to the `phpstan-baseline.neon` file?

coreylafferty:
PR to drupal-project already created/merged. This should be ready to go once conflicts resolved.

kmonahan (approved):

kmonahan:
@coreylafferty As discussed on Friday, we need both this PR and https://github.com/forumone/gesso/pull/979 for AddAttributes Twig extension

PR #949, opened 2025-04-04
911: Update nav labels to use aria-label instead of a hidden heading

### LINKED ISSUE #911: Update menu labeling
## Background
Gesso's default `nav` layout and menu block integration use a visually-hidden h2 to label the menu. However, a couple recent accessibility reviews have questioned whether the h2 truly functions as a heading and, more to the point, can result in either redundant text read out loud by a screen reader (e.g. `Main navigation, Navigation region. Main Navigation region.` if the heading is "Main navigation") or an unclear heading when navigating by the list of headings (e.g. a heading that says only "Main" but is for the main navigation).

## Work Needed
- Update relevant templates to use `aria-label` instead of an `h2` if title is hidden
- Determine recommended approach to labeling menus and implement changes as needed. Is using the block title still the right approach here?

dcmouyard:
Fixes #911.


source/03-components/breadcrumb/breadcrumb.twig
```diff
@@ -5,28 +5,11 @@
...
-  'c-breadcrumb__title',
-  title_modifier_classes ? title_modifier_classes : '',
-  hide_title ? 'u-visually-hidden' : '',
-]|join(' ')|trim %}
-
-{% set breadcrumb_id = 'breadcrumb-label'|clean_unique_id %}
-
 <nav {{ add_attributes({
-  'aria-labelledby': breadcrumb_id,
+  'aria-label': label|default('Breadcrumb'),
```
kmonahan: Let's make this `label|default('Breadcrumb'|t)` so that if no custom label is passed in, it's still translateable. Or you could remove the default, since the Drupal template is passing in a label (that is translateable).

kmonahan (changes requested):
One teensy-tiny comment. Otherwise this looks great.

source/03-components/breadcrumb/breadcrumb.twig
dcmouyard: I went with removing the default since Drupal is already passing in a translatable label.

coreylafferty (approved):

PR #951, opened 2025-04-04
928: Add stylelint-fix npm command

### LINKED ISSUE #928: Script to run eslint --fix
This is just an idea for a bit add a convenience to gesso. I've been tasked with updating the Node version on some projects, which requires updating eslint related packages. This means changes in the linting rules, and requires that I make numerous updates to sass files to adhere to those rules. Fortunately, eslint provides a command to automatically apply a portion of linting fixes where it can. The command usage is a bit tricky through, and took me a bit of time to figure out (I even gave up at one point and manually fixed several issues before going back and trying to get it to work again). Here is an example.

 ```
    stylelint -f verbose --fix \"source/**/*.scss\" --ip \"source/**/*.artifact.scss\
```

It took me a bit to figure out that I need the `--fix` part before the file  references though. I propose we simplify running this, and add it as an available  command in package.json like this: 

``` 
   "stylelintfix": "stylelint -f verbose --fix \"source/**/*.scss\" --ip \"source/**/*.artifact.scss\"",
```

You can then run `ddev gesso npm run stylelintfix`, but with a corresponding fix in `.ddev/commands/gesso`, we can make it even more convenient:

```
  stylelintfix)
    npm run stylelintfix
  ;;
```

So with this, you can just run `ddev gesso stylelintfix`

Any thoughts on this suggesion?

dcmouyard:
Fixes #928.


kmonahan (commented):
Does this also automatically fix prettier issues? If not, wonder if we should also add something like `"prettier-fix": "prettier \"source/**/*.scss\" --write"` and/or a command to run both.

PR #952, opened 2025-04-07
930: Replace aside element in sidebar layout

### LINKED ISSUE #930: Research nesting Aria landmarks
QA recently flagged (from the output of Axe Devtools) the `<aside>` in Gesso's sidebar layout being nested inside the `<main>` landmark.  Apparently `<aside>` is a complementary landmark, and should be at the top level.

https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/

We should reevaluate our use of landmarks in our layouts/templates and decide on a better approach that doesn't nest elements that should be at the top level.

dcmouyard:
Fixes #930.


kmonahan (approved):
Huh, didn't know that was disallowed. Another thing I've learned this morning!

PR #957, opened 2025-04-09
Fix button group z-index issues

dcmouyard:
This fixes edge cases when styling the focus of buttons within a button group.


kmonahan (approved):
🌮

PR #864, opened 2024-05-20
816: Use logical properties

dcmouyard:
Fixes #816. This PR:

- Switches to use CSS logical properties where applicable.
- Adds inline-start/end mixins and helper classes.

PR #958, opened 2025-04-09
Update to logical properties, removed unused CSS

dcmouyard:
Fixes a few cases that were missed in #864.


source/01-global/00-colors/color.scss
```diff
@@ -132,7 +132,7 @@ $easing-swatch: rem(75px);
   margin-block-end: 0.5rem;
   position: relative;
   transition-duration: 1s;
-  transition-property: left;
+  transition-property: inset-inline-start;
```
kmonahan: Not really an issue for this PR but why do we have easing styles in color.scss in the first place? 🤔

kmonahan (approved):
🦕

source/01-global/00-colors/color.scss
coreylafferty: Looks like we (or maybe me?) accidentally left some of the code in the color file when splitting it into multiple files.  I've removed the unused code.

PR #959, opened 2025-04-10
Fix string conversion in preprocess_menu

dcmouyard:
Fixes an edge case when first installing the theme where the url in `preprocess_menu` is a string.


kmonahan (approved):
🦑

PR #960, opened 2025-04-10
Fix responsive table script

dcmouyard:
Fixes an edge case where multiple wysiwyg fields on the same page have tables.


kmonahan (approved):
🐠

PR #962, opened 2025-05-01
Refactor scrollbar width

dcmouyard:
This PR replaces the old JS method of calculating the scrollbar width using a CSS-only approach.


kmonahan (approved):
Oooh, fancy. 😍 

I debated about the CSS property name and wrote out a whole paragraph about how I thought we should keep the non-namespaced version before deleting it because 1) there's no real downside other than "KJ has to type out more characters" and 2) it's definitely not worth holding up this PR for.

PR #968, opened 2025-07-03
added global library as a dependency to other libraries

coreylafferty:
Something has changed in recent versions of Drupal which has affected the CSS load order of libraries. In prior versions all of our library CSS was loaded after styles.css, but now many of them are loading before.  To fix this I've added the global library as a dependency on all libraries that have custom CSS.

### DIFF
--- a/gesso.libraries.yml
+++ b/gesso.libraries.yml
@@ -29,6 +29,7 @@ accordion_step_list:
     theme:
       dist/css/accordion--step-list.css: {}
   dependencies:
+    - gesso/global
     - gesso/accordion
 
 back_to_top:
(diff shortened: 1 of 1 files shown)


dcmouyard (approved):
@coreylafferty I approve of your changes.

i added a commit to remove redundant dependencies. If some of the libraries get deleted that might mess things up, so I’m fine with reverting my commit to make all of the dependencies explicit.

coreylafferty:
@dcmouyard I split the difference and added back the explicit calls to gesso/global, but left the removal of core/drupal and gesso/common (since those are part of gesso/global).  I agree the gesso/global isn't needed for libraries that depend on other libraries that call global, but I think it'd be easy to forget to add it back if you removed/deleted a dependency.

PR #979, opened 2025-10-28
Fix AddAttributes Twig extension

dcmouyard:
This PR fixes an edge case where some contrib modules pass the attributes as an array.


kmonahan (approved):
Just hit this on P'5 and can confirm Dan's fix resolved it 🙌

PR #980, opened 2025-11-17
[898] Update component script

### LINKED ISSUE #898: Update inquirer
We're currently using inquirer 9.3.6. inquirer 10 is available but even version 10 is now a legacy version and the package has been rewritten as https://www.npmjs.com/package/@inquirer/prompts. We're using inquirer for the `npm run component` script; we might also want to move to `plop` as Gesso for React now uses.

kmonahan:
Closes #898 . Updates `inquirer` to the latest version of the library, now called `@inquirer/prompts`, and rewrites the component script to use the new library. Most of these changes mirror https://github.com/forumone/nextjs-project/pull/208, but the Gesso for Drupal version includes some additions specifically for Drupal.

## New Features
- You can now generate a starting JS file as part of your component. Skeleton file imports Drupal and once and sets up Drupal behaviors
- Sass partials (files starting with `_` are kept in alphabetical order when added to the nearest `_index.scss` file)
- If your component includes a non-partial Sass file (doesn't start with `_`), and/or a JS file, the script will add the library to the libraries.yml file and add the `attach_library()` Twig function to your Twig file
- Skeleton templates are now mustache files in the `/lib/templates` directory instead of being embedded in the script itself. This comes from Gesso for React, where we were originally using Plop, and makes them easier to tweak

## Notes and Caveats
- To update the libraries YAML file, the `yaml` library converts it from YAML to JSON and back. This messes with the formatting a bit (it doesn't like extra whitespace between libraries) and strips comments
- The alphabetical ordering is handled by [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare). It doesn't always alphabetize things like I would. For example it orders `button-group, button` instead of `button, button-group`. Future work could be to do some custom sorting, but for now, it handles the major concern of not letting your index files or library file become a field of chaos.


coreylafferty (changes requested):
@kmonahan I noticed three issues and created a PR to fix two of them.  Went ahead and fixed the first one on the same branch.

1. JS files were being pointed to a 'cs' directory.  Changed to 'js'.
2. In generated story files, there was often extra line breaks between lines. My PR adds a step to remove any double spacing and replace with a single line break.
3. The new script was mangling the gesso.libraries.yml file quite a bit when adding a new library.  For example it: removed all the line breaks between libraries, removed comments, and broke some lines by changing things like `{ type: external }` to be  on a new line and removed the curly braces. My PR uses a different approach to the addition of text in the libraries file, leaving existing libraries alone.  

Note that most of this PR was written by Cursor.  I've tested it and it works, but there may be more elegant ways to write the code or issues with it I did not catch. 

https://github.com/forumone/gesso/pull/990

kmonahan:
@coreylafferty Did you see the "Notes and caveats" section? The whitespace and comments are not something I'm particularly worried about. Moving `type: external` to its own line without curly braces is just another way of writing the same thing.
```
type:
  external
```
and `{ type: external }` are identical. The latter is just a shorthand. Were you seeing something in Gesso that indicated it was broken (error messages, file not loading)?

So I am not entirely convinced we need to change the YAML file functionality. Your PR seems to add a bunch of hard-coded whitespace and treats YAML as strings rather than structured data, which I don't entirely love, since we already have a library to move between YAML and JSON.

I will take a look next week at your PR for the story spacing. I don't think extra white space gets flagged by our linters, but I'm not opposed to cleaning it up using your `normalizeStorySpacing` function.

kmonahan:
@coreylafferty Another option would be to yank the library creation part out of this PR, since I'm going to have to change it anyway for 5.5 when we move to Single Directory Components.

coreylafferty:
@kmonahan I've updated PR #990 to only change the extra spacing in story files.  

Per our discussion, please look at changing this PR so it does not alter the existing libraries file content, other than appending the new library to the end of the file.

kmonahan:
@coreylafferty Updated to add new library to end of YAML file, without reordering existing libraries or removing comments.

kmonahan:
Made one more change here, adapting https://github.com/forumone/nextjs-project/pull/230/changes#diff-8c19c8ba6bc2d5cc7e86a70206e79336519b780e70e064bf35c1a4edb7bcc639 for Gesso for Drupal. This adds another method of running the component script using command-line arguments instead of the interactive prompts. That enables AI tools to run the script as well, e.g. as part of a component creation skill.

coreylafferty (approved):

PR #987, opened 2026-01-09
806 php linting tweaks 

coreylafferty:
I ran into some additional phpstan errors when working with these changes on a real project, so I make some tweaks here.


kmonahan (approved):
I think this is fine for now. Per https://github.com/forumone/gesso/issues/945, I'm inclined to remove these commands entirely, but in the mean time, good  with anything that avoids adding more difficulty to projects.

PR #991, opened 2026-01-22
Drupal11 updates

coreylafferty:


includes/media.inc
```diff
@@ -47,7 +47,7 @@ function gesso_preprocess_filter_caption(array &$variables): void {
     $variables['classes'] = str_replace('align', 'u-align', $variables['classes']);
   }
 
-  $variables['node'] = FilteredMarkup::create(Html::serialize($dom));
+  $variables['node'] = Markup::create(Html::serialize($dom));
```
kmonahan, line 50: @coreylafferty FilteredMarkup does still exist in Drupal 11: https://api.drupal.org/api/drupal/core%21modules%21filter%21src%21Render%21FilteredMarkup.php/class/FilteredMarkup/11.x. I do see that it's marked as `@internal`, but so is Markup: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Markup.php/class/Markup/11.x. What prompted the change here?

includes/navigation.inc
```diff
@@ -41,12 +42,18 @@ function gesso_preprocess_breadcrumb(array &$variables): void {
...
     if (is_string($item['url'])) {
       $url_string = $item['url'];
     }
-    else {
+    elseif ($item['url'] instanceof Url) {
       $url_string = $item['url']->toString();
     }
+    else {
+      continue;
+    }
```
kmonahan, line 52: @coreylafferty What is the thinking behind this change? It looks like you're trying to account separately for cases where `$item['url']` isn't set and cases where `$item['url']` is set but isn't a string or a Url instance. Is that correct? If so, in the latter case, what is it?

kmonahan (commented):
Nothing that's definitely incorrect here, just some questions where I'm not quite following what you're trying to accomplish.

includes/media.inc
coreylafferty: I couldn't track down the exact error I hit that led to this change...  I think it had something to do with media captions.  I wasn't able to replicate it in my test environment when reverting this change. 

That said, I asked Cursor why this change might have been made and whether it's an improvement.  Here's what it said:

> "FilteredMarkup is @internal and intended only for the Filter module.
> For strings that have been sanitized by the filter pipeline (FilterInterface), not for arbitrary safe HTML.
> 
> In your code you’re building the string by:
> * Loading HTML with Html::load()
> * Changing the DOM (attributes, replace, etc.)
> * Serializing with Html::serialize()
> 
> That’s DOM manipulation and serialization, not “this string came from a text format / FilterInterface.” So using FilteredMarkup here is both outside its documented use and semantically wrong: you’re not saying “this was filter-processed,” you’re saying “this is safe HTML we produced.”
> 
> The right marker for “this is safe HTML, don’t escape it” in general is Markup (\Drupal\Core\Render\Markup), which the FilteredMarkup docblock explicitly points to. So the suggestion is to use Markup for your DOM-built HTML and reserve FilteredMarkup for markup that actually came from the filter system."
 
Since I can't track down the error or D11 upgrade suggestion that led me to this, I'm ok to revert it if you think the above is inaccurate and the prior code is better.

includes/navigation.inc
coreylafferty: I hit the following error in D11:

`Error: Call to a member function toString() on null in gesso_preprocess_menu() (line 48 of themes/gesso/includes/navigation.inc).`

Cursor suggested the changes here based on the error.  Here was its explanation:

> Fixed the crash by guarding against null/unexpected menu item URLs in gesso_preprocess_menu() and only calling toString() when it’s a real Url object.  
> 
> Fixed the warning by checking isset($item['url']) before touching it in gesso_preprocess_menu(); empty() still hits the array key and can warn in newer PHP.  

The change fixed the issue so I moved on, but admit this might be a case where I trusted the fix without fully understanding it.

includes/media.inc
kmonahan: As noted, they're both marked internal, so I think this is a distinction without a difference. I don't believe the change is necessary for D11 compatibility, but I don't see any harm in it, since both are doing the same thing. I'm okay leaving your change intact here.

includes/navigation.inc
kmonahan: > empty() still hits the array key and can warn in newer PHP
Hm, I'm not sure Cursor is correct about that. My understanding has long been the opposite, and that `empty()` is preferable when the item might be a falsy value OR the key might not exist at all.

I think you can simplify your solution here and do something like:
```
if (!empty($item['url'])) {
  if ($item['url'] instanceof Url) {
      $url_string = $item['url']->toString();
   }
  else {
     $url_string = $item['url'];
  }
}
```
since it sounds like we need to handle the item not having a URL at all ($item['url'] is null), it being a URL object, or it being a string.

Does that still result in any error messages?

coreylafferty: @kmonahan Added your suggestion. No errors, so I pushed it up. If this looks ok to you, I'll merge the PR in.

kmonahan (approved):

PR #1002, opened 2026-03-03
Fixes annotations for gesso link field formatters

tomealter:
Discovered this issue when attempting to render a link field using the gesso button field formatter and found that nothing was being rendered on the front end.

Drupal’s real render element annotation is @RenderElement from Drupal\Core\Render\Annotation\RenderElement. So from Drupal’s perspective, @RenderElementBase("gesso_button") is just a docblock comment, not a real plugin definition.

This PR was created in conjunction with the PR made to gesso-uswds [here](https://github.com/forumone/gesso-uswds/pull/441)


coreylafferty (approved):

gesso_helper/src/Element/GessoIcon.php
```diff
@@ -3,11 +3,12 @@
 namespace Drupal\gesso_helper\Element;
 
 use Drupal\Core\Render\Element\RenderElementBase;
+use Drupal\Core\Render\Element\RenderElement;
```
dcmouyard, line 6: We don’t need this line since we never use `RenderElement` in this file and Drupal automatically imports annotation classes.

dcmouyard (changes requested):
We can remove the unnecessary `use` statements for Classes we don’t use.

tomealter:
@dcmouyard I removed the unnecessary import in `GessoIcon`

coreylafferty (approved):

PR #1033, opened 2026-06-10
switched height and width to logical properties

coreylafferty:
Updated relevant height/width properties to block/inline size.

### DIFF
--- a/source/00-config/mixins/_accessibility.scss
+++ b/source/00-config/mixins/_accessibility.scss
@@ -6,11 +6,11 @@
 // @see http://snook.ca/archives/html_and_css/hiding-content-for-accessibility
 // @param {Boolean} $important
 @mixin visually-hidden($important: false) {
+  block-size: 1px iff($important, !important);
   clip: rect(1px, 1px, 1px, 1px) iff($important, !important);
-  height: 1px iff($important, !important);
+  inline-size: 1px iff($important, !important);
   overflow: hidden iff($important, !important);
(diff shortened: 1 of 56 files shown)


tomealter:
@coreylafferty I also found these properties that can be updated:

```1 | source/03-components/dropbutton/dropbutton.scss | 60 | width → inline-size
2 | source/03-components/accordion/accordion.scss | 37 | width → inline-size
3 | source/03-components/mobile-menu/mobile-menu.scss | 82 | width → inline-size
4 | source/03-components/dropdown-menu/dropdown-menu.scss | 96 | width → inline-size
5 | source/03-components/fieldset/_fieldset.scss | 25 | width → inline-size
6 | source/03-components/form-item/_form-item.scss | 38 | min-width → min-inline-size
7 | source/03-components/form-item/_form-item.scss | 128 | width → inline-size
8 | source/03-components/form-item/form-item--range/_form-item--range.scss | 121 | width → inline-size
9 | source/01-global/00-colors/color.scss | 20 | max-width → max-inline-size
10 | source/01-global/normalize/_normalize.scss | 88 | max-width → max-inline-size
11 | source/01-global/html-elements/24-table/_table.scss | 19 | min-width → min-inline-size
12 | source/01-global/html-elements/24-table/_table.scss | 71 | width → inline-size```

tomealter (changes requested):
Found additional items that can be udpated.

PR #1034, opened 2026-06-10
stop aggregating theme js

coreylafferty:
I keep hitting issues on projects where Drupal's aggregation of JS causes things to break (but everything works fine without aggregation).  Cursor suggested adding preprocess: false to the affected file... but also pointed out that Webpack is already minimizing the files and aggregation is redundant to prone to causing issues. Given that, I've started disabling aggregation for all our processed JS files.  This PR does the same.  Do you think we should adopt this approach?

### DIFF
--- a/gesso.libraries.yml
+++ b/gesso.libraries.yml
@@ -227,12 +227,7 @@ progress:
 
 react:
   js:
-    # Setting preprocess to false is important since we're using
-    # code-spliting (lazy()) in our React code. If Drupal's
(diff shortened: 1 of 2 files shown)


kmonahan:
Not aggregating our already minified theme JS makes sense to me. It could also let us identify more scripts as `async` or `defer`, which I should think about more than I do.

My only question is whether this particular hook approach will still work when we move to Single Directory Components, but since we've not discussed how we're testing that work or how we'd roll it out, I'm okay with that being a later problem.

PR #1035, opened 2026-06-10
moved to use null instead of false where appropriate

coreylafferty:


tomealter:
@coreylafferty why the switch to `null` for these? In twig don't `null` and `false` both produce the same falsy result in these use cases?

coreylafferty:
> @coreylafferty why the switch to `null` for these? In twig don't `null` and `false` both produce the same falsy result in these use cases?

In most cases, it won't matter. But technically false is a boolean and we typically don't mean that the variable is a boolean value... we mean that it has no value. I left the cases where it's a true boolean as 'false'.  As I recall, this came out of a discussion a while back about an edge case where JS was pulling the value from Twig and there was a minor difference between null and false. We decided in Slack that "null" was the best choice in these cases (other options considered were "false" or just "").

kmonahan (approved):

kmonahan:
> > @coreylafferty why the switch to `null` for these? In twig don't `null` and `false` both produce the same falsy result in these use cases?
> 
> In most cases, it won't matter. But technically false is a boolean and we typically don't mean that the variable is a boolean value... we mean that it has no value. I left the cases where it's a true boolean as 'false'. As I recall, this came out of a discussion a while back about an edge case where JS was pulling the value from Twig and there was a minor difference between null and false. We decided in Slack that "null" was the best choice in these cases (other options considered were "false" or just "").

Oh right, right, I remember this discussion. Yeah, in practice, it makes no difference 99.9% of the times, since `null` is false-y, as is "", but semantically, `null` better indicates "there is no thing here" and in the 0.1% of the time, if TypeScript gets involved, `null`, `false`, and `""` are all technically different types.

PR #1037, opened 2026-06-10
updated twig cache busting

coreylafferty:
Switching to use Twig Tweak's cache_metadata vs explicitly re-rendering the entire content variable.

### DIFF
--- a/templates/paragraph/paragraph--call-to-action.html.twig
+++ b/templates/paragraph/paragraph--call-to-action.html.twig
@@ -14,8 +14,4 @@
   'link_title': content.field_cta_link|field_value is not empty ? content.field_cta_link.0['#title'] : false,
 } %}
 
-{#
-  field_cta_link is intentionally omitted from the |without list because specific values are
-  extracted from it, bypassing the normal render process.
-#}
-{% set catch_cache = content|without('field_cta_media', 'field_cta_heading', 'field_cta_body') %}
+{{ content|cache_metadata }}
--- a/templates/paragraph/paragraph--hero.html.twig
+++ b/templates/paragraph/paragraph--hero.html.twig
@@ -20,9 +20,5 @@
   'hero_button_url': content.field_hero_link|field_value is not empty ? content.field_hero_link.0['#url'] : false,
 } %}
 
-{#
-  field_hero_link and field_hero_aligment are intentionally omitted from the
-  |without list because specific values are extracted from them, bypassing the
-  normal render process.
-#}
-{% set catch_cache = content|without('field_hero_image', 'field_hero_heading', 'field_hero_body') %}
+{{ paragraph|cache_metadata }}
+{{ content|cache_metadata }}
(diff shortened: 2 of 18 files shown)


kmonahan:
Does `cache_metadata` render cache on individual fields as well as on the content array itself?

kmonahan:
> Does `cache_metadata` render cache on individual fields as well as on the content array itself?

Checking Slack and we settled this in 2025: `|cache_metadata` is the better approach 🎉

templates/paragraph/paragraph--call-to-action.html.twig
```diff
@@ -14,8 +14,4 @@
   'link_title': content.field_cta_link|field_value is not empty ? content.field_cta_link.0['#title'] : false,
 } %}
 
-{#
-  field_cta_link is intentionally omitted from the |without list because specific values are
-  extracted from it, bypassing the normal render process.
-#}
-{% set catch_cache = content|without('field_cta_media', 'field_cta_heading', 'field_cta_body') %}
+{{ content|cache_metadata }}
```
kmonahan, line 17: Per the original code comment above, I think we also need `{{ content.field_cta_link|cache_metadata }}` to account for the fact that `field_cta_link` isn't being rendered normally.

templates/paragraph/paragraph--hero.html.twig
```diff
@@ -20,9 +20,5 @@
...
 } %}
 
-{#
-  field_hero_link and field_hero_aligment are intentionally omitted from the
-  |without list because specific values are extracted from them, bypassing the
-  normal render process.
-#}
-{% set catch_cache = content|without('field_hero_image', 'field_hero_heading', 'field_hero_body') %}
+{{ paragraph|cache_metadata }}
+{{ content|cache_metadata }}
```
kmonahan, line 23: Same as above. I'm not sure we need `{{ paragraph|cache_metadata }}` since `paragraph` isn't a render array; it's the Paragraph object, so any cache metadata on it should already be covered by `content`. But we may need the `field_hero_link` and `field_hero_alignment` metadata since we're extracting specific values from them instead of rendering them normally.

kmonahan (changes requested):
I think when we are bypassing the render array entirely for a field, as we do some times with links, we still need to render their cache_metadata explicitly. Previously, that was handled by omitting them from the `|without()` filter. Otherwise, hooray for not having to write long lists of field names or face side effects!

Also well done on the branch name 😂

templates/paragraph/paragraph--call-to-action.html.twig
coreylafferty: I don't think we need to call out the cta_link field specifically since content|cache_metadata will include it along with all other fields.

templates/paragraph/paragraph--hero.html.twig
coreylafferty: I went ahead and removed the paragraph call.  For the other fields, I think content|cache_metadata will cover them. Ran this by Cursor, and (for what it's worth) it agrees.

kmonahan: I'm looking at https://git.drupalcode.org/project/twig_tweak/-/blob/4.x/src/CacheMetadataExtractor.php?ref_type=heads but it's hard to tell. It is doing something with the children of the render array, because it's calling `$keys = Element::children($build);`, so fine for now.

kmonahan (approved):

PR #1043, opened 2026-07-09
Replace theme_get_setting() with a D10/D11-compatible ThemeSettings h…

coreylafferty:
…elper.

### DIFF
--- a/gesso_helper/src/ThemeSettings.php
+++ b/gesso_helper/src/ThemeSettings.php
@@ -14,10 +14,10 @@ namespace Drupal\gesso_helper;
 class ThemeSettings {
 
   /**
-     * The theme settings provider, when available.
-     *
-     * @var object|null
-     */
+   * The theme settings provider, when available.
+   *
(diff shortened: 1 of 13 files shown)


gesso_helper/src/ThemeSettings.php
```diff
@@ -0,0 +1,55 @@
...
+namespace Drupal\gesso_helper;
+
+/**
+ * Provides cross-version theme settings access.
+ *
+ * Uses ThemeSettingsProvider on Drupal 11.3+, with a theme_get_setting()
+ * fallback for Drupal 10.3 through 11.2.
+ *
+ * The constructor argument is intentionally untyped: ThemeSettingsProvider
+ * does not exist before Drupal 11.3, so a class typehint would fatal on load.
```
kmonahan: Super minor note: there's a typo here. Should be "so a class typehint would **be** fatal on load."

kmonahan (approved):

PR #1060, opened 2026-08-05
Single directory components cl tweaks

coreylafferty:
* **Library / CSS load order** 
  * Move style.css to base so it loads before components
  * attach button CSS on form buttons
  * fix broken library paths and drop dead attach_library calls
  * restore mobile-menu/hamburger deps on menus
* **Cross-component CSS deps** 
  * Explicit imports so form-item variants, responsive tables, menus, etc. get their CSS/JS when needed
* **SDC schemas / attributes** 
  * Consistency sweep: add/tighten attributes, slots, props
  * Views vars/props fixed to stop crashes
* **Breadcrumb split** 
  * Real breadcrumb component + styles vs region wrapper (breadcrumb-wrapper)
* **Page title**
  * Bring back page-title block to keep the title component simple
* **Button-group-item** 
  * Refactor to match accordion-item (child SDC pattern)
* **Component generator (lib/component.js)** 
  * Overhaul templates/README so new SDCs get correct yml/stories/scripts
* **Storybook** 
  * Aliases, story imports for asset loading, image_path, editor CSS option, lint/namespace cleanup
* **Templates / edge cases** 
  * Maintenance page guards, hamburger/menu attributes, pager tweaks, minor page template fixes.

TL;DR: Make SDCs actually work in Drupal (libraries, attrs, Views) and Storybook (deps, stories), not just sit in folders like unused Xenomorph eggs.


.storybook/getArgTypesFromComponent.js
```diff
@@ -1,39 +1,62 @@
...
     ...Object.fromEntries(
-      Object.entries(componentInfo.props.properties || {}).map(([key, p]) => {
+      Object.entries(componentInfo.props?.properties || {}).map(([key, p]) => {
         const argTypeConfig = {};
         if (p.title) {
           argTypeConfig.name = p.title;
         }
-        if (p.type) {
-          argTypeConfig.type = p.type;
+        const storybookType = normalizeStorybookType(p.type);
```
kmonahan, line 26: Did you run into an issue where the type wasn't normalized? If so, where?

.storybook/main.js
```diff
@@ -86,14 +87,9 @@ const config = {
...
-                components: resolve(__dirname, '../', 'source/03-components'),
-                templates: resolve(__dirname, '../', 'source/04-templates'),
-                pages: resolve(__dirname, '../', 'source/04-pages'),
-                // SDC (Single Directory Component) namespace
-                // Supports Drupal's SDC syntax: include('gesso:component-name')
-                // Components in the components/ directory follow the convention:
-                // gesso:page-title resolves to components/page-title/page-title.twig
+                utility: resolve(__dirname, '../', 'source/02-utility'),
+                layouts: resolve(__dirname, '../', 'components/layouts'),
+                components: resolve(__dirname, '../', 'components/components'),
```
kmonahan: Why was this necessary? We don't put Twig in our `utility` directory (by design) and so it wasn't included in the original mapping. 

For `layouts` and `components`, those wouldn't be the namespaces, would they? I would have thought it was still `gesso`. Do you remember what story was broken that prompted this change?

components/components/accordion-item/accordion-item.component.yml
```diff
@@ -1,8 +1,15 @@
...
 props:
   type: object
   properties:
+    attributes:
+      type:
+        - object
+        - string
+        - "null"
+      title: Attributes
+      description: 'HTML attributes for the root element.'
```
kmonahan: Looking at https://www.drupal.org/docs/develop/theming-drupal/using-single-directory-components/annotated-example-componentyml as an example, `attributes` is not included, and I don't think it's something we would need to specify as prop. Did you see something that wasn't working with `attributes` not listed? If not, you're right that this should be consistent, but I think it needs to be consistent in the other direction. I can have CC take a pass and remove anywhere `attributes` is listed.

components/components/back-to-top/back-to-top.twig
```diff
@@ -9,5 +9,6 @@
     'label': text,
     'modifier_classes': 'c-icon--back-to-top',
     'icon_name': 'angle',
+    'image_path': image_path,
```
kmonahan: Hm, I'm surprised this was needed, since we're not using `only`. We'll need to keep an eye on whether SDC changes how templates inherit variables that aren't passed explicitly.

components/components/filter/filter.twig
```diff
@@ -13,6 +14,7 @@
     'modifier_classes': 'c-icon c-button__icon is-spaced-after',
     'label': 'Remove'|t,
     'icon_name': 'xmark',
+    'image_path': image_path,
```
kmonahan: I'm noticing `image_path` is getting explicitly passed in a lot. Were all of these not working until `image_path` was passed in? If so, continuing my earlier comment, that's a behavior change with SDC will want to be aware of.

components/components/form-item/form-item--checkbox/_form-item--checkbox.scss
```diff
@@ -1,7 +1,7 @@
 // Component: Checkbox Form Item
 
 @use '00-config' as *;
-@use '../form-item' as *;
+@use '../form-item-vars' as *;
```
kmonahan: Good idea. This is definitely cleaner than importing the base form-item styles every time.

components/components/form-item/form-item--checkboxes/form-item--checkboxes.twig
```diff
@@ -1,3 +1,4 @@
+{{ attach_library('core/components.gesso--form-item') }}
```
kmonahan: Would adding the library as a dependency in the YAML file accomplish the same thing? If so, we should probably standardize on one approach versus the other.

components/components/form-item/_input.twig
```diff
@@ -1,3 +1,4 @@
+{{ attach_library('core/components.gesso--form-item') }}
```
kmonahan: Is the concern here that `_input` would be used outside of `form-item` and so the general form-item library wouldn't be attached? I don't think partials should bring all their component's styling with them generally, but willing to reconsider if there's a particular use case you're trying to cover.

components/components/form-item/form-item.source.scss
```diff
@@ -2,14 +2,15 @@
...
-$form-text-size: gesso-base-font-size() !default;
+@use 'form-item-vars' as *;
+
+// Variant styles (underscore partials are ignored by webpack entry globs).
+@use 'form-item--checkbox/form-item--checkbox';
+@use 'form-item--checkboxes/form-item--checkboxes';
+@use 'form-item--radio/form-item--radio';
+@use 'form-item--radios/form-item--radios';
+@use 'form-item--range/form-item--range';
+@use 'form-item--select/form-item--select';
```
kmonahan: Okay, looking at this and some of my earlier comments, I think what needs to happen here is separating out which form-item subitems are components in their own right and which can't be used outside of form-item. I'm okay with going forward with what you have here and potentially doing some cleanup in Gesso 6.0.1

components/components/hamburger-button/hamburger-button.twig
```diff
@@ -3,6 +3,6 @@
   modifier_classes ? modifier_classes : ''
 ]|join(' ')|trim %}
 
-<button {{ add_attributes({class: classes}) }}>
-  <span class="c-hamburger-button__icon">Menu</span>
+<button type="button" {{ add_attributes({class: classes}) }}>
```
kmonahan: This should be in the object that's passed to `add_attributes`, since it's an attribute

components/components/views-view-list/views-view-list.component.yml
```diff
@@ -62,27 +58,21 @@ props:
...
-    row_attributes:
-      type:
-        - object
-        - "null"
-      title: Row Attributes
-      description: 'HTML attributes for row items.'
+    # `rows` is intentionally not declared as a prop. Each row's `content` is a
+    # Drupal render array that often contains circular references; declaring it
+    # causes ComponentValidator json_encode recursion fatals. Still passed from
+    # Twig / Storybook as a free variable: [{ content, attributes? }, ...].
```
kmonahan, line 54: Huh, interesting. Cool, let's leave it out for now and see how a future Drupal release handles it.

components/layouts/breadcrumb-wrapper/breadcrumb-wrapper.source.scss
```diff
@@ -0,0 +1,3 @@
+// Layout: Breadcrumb Wrapper
+
+// .l-breadcrumb-wrapper {}
```
kmonahan: This will compile to an empty CSS file, which will still get attached, won't it? I think we should remove it if that's the case, and someone who needs breadcrumb-wrapper styling can add it then.

includes/form.inc
```diff
@@ -25,9 +25,12 @@ function gesso_preprocess_input(array &$variables): void {
   $variables['type'] = $variables['element']['#attributes']['type'] ?? NULL;
   $variables['dropbutton'] = $variables['element']['#dropbutton'] ?? NULL;
 
-  // Replace button classes with c-button.
+  // Replace button classes with c-button and attach button CSS when needed.
```
kmonahan, line 28: Ah, good thought!

lib/component.js
```diff
@@ -4,25 +4,17 @@
...
-} from 'node:fs/promises';
+import { access, mkdir, readFile, writeFile } from 'node:fs/promises';
 import path from 'node:path';
 import { EOL } from 'node:os';
-import { parseDocument } from 'yaml';
 
-const __dirname = import.meta.dirname;
-const sassPathMatcher = RegExp(/@use (['"]{1})(?<path>.+)\1/);
+/** Allowed SDC output folders under components/. */
+const VALID_FOLDERS = ['components', 'layouts', 'templates'];
```
kmonahan: I think this should be dynamic as the previous version was. Any subfolder under `components` should be a valid location.

lib/component.js
```diff
@@ -117,12 +104,11 @@ function validateRequiredArgs(parsed) {
...
  */
-function validateFolder(folder, validFolders) {
-  if (!validFolders.includes(folder)) {
+function validateFolder(folder) {
+  if (!VALID_FOLDERS.includes(folder)) {
     console.error(`Error: Invalid folder "${folder}".`);
-    console.error(`Valid options: ${validFolders.join(', ')}`);
+    console.error(`Valid options: ${VALID_FOLDERS.join(', ')}`);
     process.exit(1);
   }
```
kmonahan, line 117: I think this should remain as it was

lib/component.js
```diff
@@ -133,64 +119,16 @@ function validateFolder(folder, validFolders) {
...
-
-/**
- * Checks whether the source directory is an accessible directory.
- * @param {node:fs.PathLike} source - Source path
- * @return {Promise<boolean>} - True if source is an accessible directory
- */
-async function isDirectory(source) {
-  const stats = await lstat(source);
-  return stats.isDirectory();
-}
```
kmonahan, line 164: No reason this needs to be removed?

lib/component.js
```diff
@@ -133,64 +119,16 @@ function validateFolder(folder, validFolders) {
...
-          '02-utility',
-          'fonts',
-          'images',
-        ].includes(dirName)
-    )
-    .map(name => path.join(source, name));
-  /** @type {Awaited<boolean>[]} */
-  const isDirectoryResults = await Promise.all(directoryPaths.map(isDirectory));
-  return directoryPaths.filter((value, index) => isDirectoryResults[index]);
-}
```
kmonahan, line 191: This should be updated, not removed

lib/component.js
```diff
@@ -240,18 +176,33 @@ async function getComponentFolder() {
...
+ */
+function storySection(directory) {
+  if (directory === 'layouts') {
+    return 'Layouts';
+  }
+  if (directory === 'templates') {
+    return 'Templates';
+  }
+  return 'Components';
+}
```
kmonahan: This doesn't seem necessary. We can use the directory and convert the case.

lib/component.js
```diff
@@ -457,34 +349,20 @@ async function generator() {
...
-    addAttributes: `{{ add_attributes({ class: classes }) }}`,
-    attachLibrary: `{{ attach_library('gesso/${componentName}') }}`,
+    addAttributes: `{{ add_attributes({ 'class': classes }) }}`,
     // Lambdas
-    machineName: (text, render) => {
-      return pascalCase(render(text));
-    },
-    humanName: (text, render) => {
-      return camelCase(render(text));
-    },
```
kmonahan, line 474: These should stay. No reason to remove them I know of

lib/component.js
```diff
@@ -457,34 +349,20 @@ async function generator() {
...
     pascalCase: (text, render) => {
       return pascalCase(render(text));
     },
     camelCase: (text, render) => {
       return camelCase(render(text));
     },
-    titleCase: (text, render) => {
-      return capitalCase(render(text));
-    },
```
kmonahan, line 486: This should stay.

templates/block/block--system-branding-block.html.twig
```diff
@@ -17,6 +17,7 @@
 {% block block_content %}
   {% if site_logo %}
     {% include 'gesso:site-logo' with {
+      'attributes': null,
       'url': path('<front>'),
       'site_logo': site_logo,
     } %}
```
kmonahan, line 23: Hm, this seems at-odds with all of the `image_path` changes. Why are attributes still silently passed down but image_path isn't?

templates/facets/facets-result-item--summary.html.twig
```diff
@@ -14,7 +14,8 @@
  *   - label: The facet label.
  */
 #}
-{% include '@gesso/components/components/filters/filter.twig' with {
+{% include '@components/filters/filter.twig' with {
```
kmonahan: The original is wrong but so is the correction. It should be `gesso:filters`

templates/form/checkboxes.html.twig
```diff
@@ -11,6 +11,6 @@
  */
 #}
 {# 'form-checkboxes' is a Drupal core class used by some modules (e.g., webform) #}
-{% include '@gesso/components/components/form-item/form-item--checkboxes/form-item--checkboxes.twig' with {
+{% include '@components/form-item/form-item--checkboxes/form-item--checkboxes.twig' with {
```
kmonahan: This should stay as is, unless you saw that it wasn't actually working. OR checkboxes needs to be its own component, which might actually be the correct move for here and elsewhere, if we're importing the item directly into a template. That goes along with my previous comments about needing to straighten out form-item.

gesso.info.yml
```diff
@@ -25,6 +23,7 @@ components:
   namespaces:
     global: source/01-global
     layouts: components/layouts
+    components: components/components
     utility: source/02-utility
```
kmonahan, line 27: This entire section should be removed, so then we can remove the dependency on Components

gesso.libraries.yml
```diff
@@ -4,13 +4,21 @@ common:
 
 global:
   css:
+    # Base so SDC component CSS (weight 0) loads after global resets/element
+    # styles (weight -200). Declaring gesso/global as an SDC dependency is not
+    # enough — see https://www.drupal.org/project/drupal/issues/3560103
```
kmonahan, line 9: Oof, well that's a fun bug.

gesso.libraries.yml
```diff
@@ -4,13 +4,21 @@ common:
...
+      dist/css/styles.css: {}
     theme:
-      # If you alter the fonts here, be sure to also update: source/ckeditor4-styles.scss, source/editor-styles.scss and .storybook/preview-head.html.
+      # If you alter the fonts here, be sure to also update: source/editor-styles.scss and .storybook/preview-head.html.
       //fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,300;0,400;0,600;0,700;1,300;1,400;1,600;1,700&display=swap: { type: external }
-      dist/css/styles.css: {}
+      # Table CSS/JS also live on gesso:table. Kept on global too so WYSIWYG
+      # (and any other .c-table markup that never renders through the SDC)
+      # still gets styles + sortable behavior.
+      components/components/table/table.css: {}
```
kmonahan: Hmm, if they're loaded globally, then I think they should live in `01-global` and *not* in a component.

package.json
```diff
@@ -13,8 +13,8 @@
     "watch-design-tokens": "webpack watch --config ./webpack.theme-config.js --mode=development",
     "watch-react": "webpack watch --config ./webpack.react-config.js",
     "watch": "npm run start && concurrently --raw \"npm:watch-theme\" \"npm:watch-design-tokens\" \"npm:watch-react\"",
-    "eslint": "eslint source/**/!\\(*.stories\\).js",
-    "stylelint": "stylelint \"source/**/*.scss\"",
+    "eslint": "eslint \"source/**/!(*.stories).{js,jsx,ts,tsx}\" \"components/**/!(*.stories).{js,jsx,ts,tsx}\" \"lib/**/*.{js,cjs}\" \".storybook/**/*.{js,jsx}\" \"stories/**/*.{js,jsx}\"",
```
kmonahan: Flagging that this is a lot more linting than we were doing before. Not a bad thing, but a change beyond switching to SDC.

kmonahan (commented):
CC and I will work on making changes and open a grandchild branch, but there are also questions for you here.

gesso.info.yml
kmonahan: We're still using components for some imports of `@global`, so leaving it in place.

.storybook/getArgTypesFromComponent.js
coreylafferty: Yes — this showed up once we started using JSON Schema union types in component.yml (e.g. type: [string, "null"]). The YAML parser gives Storybook an array, but argTypes.type expects a scalar. Without normalization, stories using getArgTypesFromComponent (article, views-view, etc.) get invalid type values in the Controls panel. The helper picks the first non-null member of the union for simple types, and sets control: false for object/array types that can't have a basic control anyway.

.storybook/main.js
coreylafferty: Removed unused paths.

.storybook/getArgTypesFromComponent.js
kmonahan: Huh, interesting. That makes good sense.

components/components/back-to-top/back-to-top.twig
coreylafferty: I think you're right -- not needed.  Removed image_path from several files in commit: https://github.com/forumone/gesso/pull/1060/changes/a51298e3045121bd51ac7dc4c8e60ea266ac9567

templates/block/block--system-branding-block.html.twig
coreylafferty: Removed most of the redundant image_path calls in response to prior comment(s).  Leaving this as-is to stop the block classes and attributes from leaking into the site-logo and such.

package.json
coreylafferty: Expanded because SDC moved JS/TS/SCSS into components/, added lib/component.js, and touched .storybook/ — the old scripts only covered source/**, so npm test would miss the new locations. Webpack was already linting components/**/*.scss (and component source JS/TS via ESLintPlugin); this brings the npm scripts in line.

kmonahan: Agree with linting `components`. I should have been more clear -- this change also adds linting to `lib` and `.storybook`, which is what prompted my comment. I don't think there's any reason we need to lint `.storybook` -- that's just Storybook config files. I'm on the fence about linting `lib` and think I'm gonna go with "let's look at that separately from the SDC conversion." Some of those files are getting a bit musty and should potentially be converted to TypeScript, so I'd rather not start linting them yet. What's most important to lint is the JS that's actually part of the theme and served client-side, which is what's in `source` and `components.`

gesso.libraries.yml
coreylafferty: I dug into this deeper and off-the-shelf, only l-responsive-table can appear in the WYSIWYG.  I simplified this and stopped loading the c-table css/js globally.  Commit: https://github.com/forumone/gesso/pull/1060/changes/ae737cf276e137fb4b5e0da64c3443264a454388

kmonahan (approved):

PR #1065, opened 2026-08-11
2026-08 package updates

kmonahan:
Nothing exciting here since we're still waiting for some other plugins to catch up before we jump to ESLint 10


tomealter (changes requested):
Looks like a handful of packages already have newer minor/patch releases available beyond current package.json ranges:

`@storybook/* (10.5.7 → 10.5.10), @swc/core (1.15.47 → 1.16.1), concurrently, html-react-parser, sass-embedded, @inquirer/prompts, @forumone/eslint-config-es5/-react`

All are semver-compatible with your existing ^ ranges, so a plain npm update would pick them up.

kmonahan:
@tomealter I do these once a month, so the next round of updates will get picked up in September. If this branch isn't merged into 5.x-RC  by then, I'll add on to this branch. Otherwise, I'll make a new one.

Please mark as changes requested if and only if you see blockers with the existing changes (e.g. the latest update broke something).

tomealter (approved):

coreylafferty (approved):

PR #1067, opened 2026-08-18
Fix React linting rules not being applied

kmonahan:
Discovered that the React-specific linting rules weren't actually being applied to JSX/TSX files.


package.json
tomealter: Do we want to update the `eslint` command to include .jsx and .tsx as well?
`"eslint": "eslint \"source/**/!(*.stories).{js,jsx,ts,tsx}\""`

tomealter (commented):
@kmonahan just one question here

package.json
kmonahan: Hmm, yeah, that seems like a good idea 👍

kmonahan:
@coreylafferty This is updated. You will need to run `ddev gesso install` or `npm ci` to make sure you have the latest version of `@forumone/eslint-config-react`. If you are testing via ddev, make sure your `eslint` command is running `npm run eslint` (same fix we made on NGA) and restart ddev if you had to change it. (Which reminds me that I also need to get a `drupal-project` PR open to fix that.)

I'm seeing all JS and SCSS linting passing on a fresh Gesso checkout with the latest here in place.

kmonahan:
> @coreylafferty This is updated. You will need to run `ddev gesso install` or `npm ci` to make sure you have the latest version of `@forumone/eslint-config-react`. If you are testing via ddev, make sure your `eslint` command is running `npm run eslint` (same fix we made on NGA) and restart ddev if you had to change it. (Which reminds me that I also need to get a `drupal-project` PR open to fix that.)
> 
> I'm seeing all JS and SCSS linting passing on a fresh Gesso checkout with the latest here in place.

drupal-project PR now opened as well: https://github.com/forumone/drupal-project/pull/222

coreylafferty (approved):

PR #1069, opened 2026-08-27
Fix missing css due to splitChunks

dcmouyard:
Ran into an issue where an import of a library stylesheet was broken (e.g. `@import 'swiper/swiper-bundle.css';`) due to the splitChunks optimization. This PR just limits the js/common cache group to just JavaScript files.

We could also add another cache group for CSS files and add it to the common library, but I’m not sure how often this would be used on projects.


kmonahan:
@dcmouyard How can I replicate your original issue? Where/how were you importing swiper?

